PR #3, opened 2015-08-12
Caching support for Dockerfile content (#1)

### LINKED ISSUE #1: RFE: Caching support for Dockerfile content
As explained in https://github.com/projectatomic/osbs-client/issues/200, it'd be great if `DockerfileParser` could cache `Dockerfile` content when instantiated.
Since I think the purpose of your library is not only to read existing Dockerfiles, but also to create new Dockerfiles, this can't be done automatically by constructor (i.e. I can create a `DockerfileParser` instance with a path to not-yet-existing Dockerfile, add the contents and then dump it).
So I think the best course of action would be to add an argument to the constructor, something like `cache_content=False`. If this is `True`, Dockerfile content would get cached and `DockerfileParser` would then operate on the cached data. With `False`, everything would work pretty much the same way as it does now.

jpopelka:
When the caching is on:

a) reading of Dockerfile content/lines:
 return cached content if there's any, otherwise read the Dockerfile (and store it to cache)

b) writing of Dockerfile content/lines:
 write to cache as well as into real Dockerfile

Fixes #1

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -18,17 +18,22 @@ from .constants import DOCKERFILE_FILENAME, PY2
 
 logger = logging.getLogger(__name__)
 
+
 class DockerfileParser(object):
-    def __init__(self, path=None):
+    def __init__(self, path=None, cache_content=False):
         """
         Initialize path to Dockerfile
         :param path: path to (directory with) Dockerfile
+        :param cache_content: cache Dockerfile content inside DockerfileParser
         """
         path = path or '.'
         if path.endswith(DOCKERFILE_FILENAME):
             self.dockerfile_path = path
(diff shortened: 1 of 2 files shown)


bkabrda:
This looks good, with one small exception - it doesn't cache the content until you actually access it. Would it be possible to add something like `self.content` (which would trigger caching) to `__init__`, so that the content actually gets cached in the constructor?
The implication of this would be that it wouldn't be possible to create parser instance for a nonexistent Dockerfile with caching on... or, the cached content for a nonexistent file would be just empty.

jpopelka:
@bkabrda check the new commit please

bkabrda:
Exactly what I need. No further suggestions from my side :)

Issue #2, opened 2015-08-12
Support for modifying labels


twaugh:
With the `structure` property it is quite easy to modify most values, by first finding the value to modify, then replacing the lines (known from `startline` and `endline`) in the `lines` property.

For example, for CMD:
```
for insn in parser.structure:
  if insn['instruction'] == 'CMD':
    content = 'CMD my-new-cmd-value\n'
    lines = parser.lines
    del lines[insn['startline']:insn['endline'] + 1]
    lines.insert(startline, content)
    parser.lines = lines
    break
```

However, for the LABEL instruction it is much more complicated as firstly there are two forms of LABEL, secondly there is quoting to deal with, and thirdly a single LABEL instruction can contain many labels.

It would be nice to provide support for modifying a single label, handling all these cases.

jpopelka:
@twaugh  Please check #4, the new code might be familiar to you ;-)

twaugh:
Closed via #4.

PR #4, opened 2015-08-14
 Support for modifying instructions (labels especially) (#2)

jpopelka:


dockerfile_parse/parser.py
```diff
@@ -236,3 +264,73 @@ def labels(self):
                         labels[key_val[0]] = key_val[1] if len(key_val) > 1 else ''
                         logger.debug("new label %s=%s", repr(key_val[0]), repr(labels[key_val[0]]))
         return labels
+
+    def modify_instruction_label(self, label_key, label_value):
+        """
+        Rewrite a local Dockerfile with an incremented Release label
```
twaugh: The doc text should probably be more generic. :-)

dockerfile_parse/parser.py
```diff
@@ -236,3 +264,73 @@ def labels(self):
                         labels[key_val[0]] = key_val[1] if len(key_val) > 1 else ''
                         logger.debug("new label %s=%s", repr(key_val[0]), repr(labels[key_val[0]]))
         return labels
+
+    def modify_instruction_label(self, label_key, label_value):
```
twaugh: I guess a more Pythonic way of doing it would be to use setters for the properties, but would that be really hard?

jpopelka: Is the additional commit what you've had in mind ?

jpopelka:
I've added real labels setter, so you have 2 options now:
  * df.labels = {'name':'value'}
- the real setter, this *sets* (i.e. remove old ones) LABELs to the specified ones
  * df.change_labels({'name':'value'})
- this only changes value of existing LABELs

I haven't managed to make the `x.labels['name'] = 'value'` work without breaking other things :-(

twaugh:
OK, never mind about setting `x.labels['name']` directly.

For `change_labels()` maybe we should raise an exception if we don't find the LABEL key, rather than failing an assertion.

I still think it would be nice for us to be able to make minimal changes to LABELs even when setting them, like in the tests I put here: https://github.com/jpopelka/dockerfile-parse/compare/modify-labels...twaugh:modify-labels-tests

It shouldn't be hard: getting two sets of LABEL keys, ones we already have and ones we need to end up with, we can easily find out which we need to add/delete/modify, with `set() - set()`-type calculations.

jpopelka:
> For change_labels() maybe we should raise an exception if we don't find the LABEL key, rather than failing an assertion.

already changed to KeyError

> I still think it would be nice for us to be able to make minimal changes to LABELs even when setting them

done


> It shouldn't be hard ...

grrr, I don't want to see quotations for a veeery long time ;-)

twaugh:
This looks really nice!

jpopelka:
Could you check the new commit please ?
Thanks, I really appreciate all your remarks.

jpopelka:
OK, I think I've lost this battle with quotes :-(
With the last commit when we change value of label which is on one line with other labels, those other labels will be in quotes even they weren't previously (which is not wrong per se).

You can revert this 'revert' commit if you want to try, but I'm really at my wit's end with this.
I don't want to spend more time with this as I think this state is good enough.

twaugh:
LGTM!

jpopelka:
Thank you Tim !

PR #5, opened 2015-08-26
Support for modifying labels

twaugh:
Here is a set of changes based on https://github.com/DBuildService/dockerfile-parse/pull/4.

I started from where we got up to on Aug 17th, fixed some of the test cases and added a couple more based on bugs I found.

I've also added a commit at the end which implements the `x.labels['foo'] = 'bar'` functionality. See what you think.


jpopelka:
> I've also added a commit at the end which implements the x.labels['foo'] = 'bar' functionality.

Oh my, that's wonderful.

I've rebased the original `modify-labels` branch, added one commit with more tests and it should be ready to finish back in #4.

PR #6, opened 2015-09-16
handle ENV similar to LABEL

jpopelka:
ENV can now be handled similarly to LABEL, allowing modifications of individual environment variables.


twaugh:
Nice work! LGTM. I don't see any regressions.

(https://github.com/DBuildService/dockerfile-parse/issues/7 happens regardless of this change.)

jpopelka:
Merged, thanks.

PR #9, opened 2015-09-17
Perform environment variable substitution by default (fixes #8)

### LINKED ISSUE #8: Should handle environment variable substitution
Docker allows environment variable substitution in a number of instructions. This parser ought to do likewise.

Additionally, it would be useful to be able to disable environment variable substitution, to inspect the unquoted-but-not-substituted values.

twaugh:
I've added a new class `EnvSubst` for doing environment variable substitution when quoting allows, as this doesn't seem to be possible/easy with other approaches I investigated (subclassing `shlex.shlex`, using `string.Template`, using `os.path.expandvar()`).

The `shlex_split()` function now has keyword arguments for performing environment substitution.

The instructions getter method now always tracks `ENV` in order to build an environment variable dictionary with which to perform substitution.

The `DockerfileParser` constructor now takes a keyword argument `env_replace` to specify whether to perform substitution. An instance variable of the same name can be set after construction for the same effect.

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -66,7 +66,7 @@ class Envs(dict):
 
 
 class DockerfileParser(object):
-    def __init__(self, path=None, cache_content=False):
+    def __init__(self, path=None, cache_content=False, env_replace=True):
         """
         Initialize path to Dockerfile
         :param path: path to (directory with) Dockerfile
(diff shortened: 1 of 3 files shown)


twaugh:
This also fixes #9.

twaugh:
Fixes #8 (not sure if having that in the title is sufficient to close it when merged).

TomasTomecek:
this is insanely impressive

twaugh:
Thanks!

twaugh:
Updated. I'm happy to squash before merge but I've left the fix-up commits as they are for review.

jpopelka:
Awesome !
Feel free to squash & merge.

PR #16, opened 2016-05-06
Skip the line in Dockerfile if it is a comment

### LINKED ISSUE #2: Support for modifying labels

vrutkovs:
Fixes #2

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -53,7 +53,9 @@ CMD {0}""".format(NON_ASCII)
                           "    # comment\n",    # should be ignored
                           "    bar  \n",        # extra ws, continuation line
                           "USER  {0}\n".format(NON_ASCII),
-                          "# comment \\\n",    # extra ws
+                          "# comment \\\n",     # extra ws
+                          "# comment \\ \n",    # extra ws with a space
+                          "# comment \\\\ \n",  # two backslashes
                           "RUN command1\n",
                           "RUN command2 && \\\n",
                           "    # comment\n",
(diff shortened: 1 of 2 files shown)


twaugh:
I would like more tests, so that we know how these sorts of things behave:

```
# ignore this \
RUN but run this
```

```
RUN run some stuff && \
    # ignore this, but \
    run some other stuff
```

For the record, I think the docker parser's behaviour here is terrible, but we have to do what it does.

tests/test_parser.py
```diff
@@ -52,7 +52,12 @@ def test_dockerfile_structure(self, dfparser):
                           " label  foo  \\\n",  # extra ws
                           "    # comment\n",    # should be ignored
                           "    bar  \n",        # extra ws, continuation line
-                          "USER  {0}".format(NON_ASCII)]   # extra ws, no newline
+                          "USER  {0}\n".format(NON_ASCII),
+                          "# comment \\\n",    # extra ws
```
twaugh: How about adding a test with extra whitespace *after* the backslash as well? (You'll be surprised what docker does with that...)

twaugh: Also how about two backslashes at the end of the line?

vrutkovs:
Implemented those, I really hope Docker's parser won't crash on those cases

twaugh:
Is this still current? I see #2 was marked as being closed by #4.

vrutkovs:
It is, it fixes several other cases (e.g. comment in multi-line instruction)

PR #19, opened 2016-10-19
README: add more examples

ktdreyer:
Fix the flake8 warnings in the README's example, and add a couple more
examples (setting a new base image, printing the resulting Dockerfile)


twaugh:
Thanks!

Issue #20, opened 2016-11-15
Should accept dict representing environment vars inherited from parent


twaugh:
When substituting environment variables, the environment is assumed to be empty at the start. This is not the case for images that inherited from anything other than `scratch`.

The constructor should accept an optional dict parameter `parent_env`, and this should be copied and used as the starting environment variable when parsing the Dockerfile.

maxamillion:
https://github.com/DBuildService/dockerfile-parse/pull/21
https://github.com/projectatomic/atomic-reactor/pull/586

twaugh:
Fixed via #21.

PR #21, opened 2016-11-16
add parent_env handling

maxamillion:
Signed-off-by: Adam Miller <maxamillion@fedoraproject.org>

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -66,11 +66,12 @@ class Envs(dict):
 
 
 class DockerfileParser(object):
-    def __init__(self, path=None, cache_content=False, env_replace=True):
+    def __init__(self, path=None, cache_content=False, env_replace=True, parent_env=None):
         """
         Initialize path to Dockerfile
         :param path: path to (directory with) Dockerfile
         :param cache_content: cache Dockerfile content inside DockerfileParser
+        :param parent_env: python dict of inherited env vars from parent image
         """
         path = path or '.'
         if path.endswith(DOCKERFILE_FILENAME):
@@ -90,6 +91,12 @@ class DockerfileParser(object):
                 pass
 
         self.env_replace = env_replace
+
+        if parent_env:
+            logger.debug("Setting inherited parent image ENV vars: %s", parent_env)
+            self.parent_env = parent_env
+        else:
+            self.parent_env = {}
 
     @property
     def lines(self):
@@ -290,7 +297,7 @@ class DockerfileParser(object):
         if name != 'LABEL' and name != 'ENV':
             raise ValueError("Unsupported instruction '%s'", name)
         instructions = {}
-        envs = {}
+        envs = self.parent_env.copy()
         for insndesc in self.structure:
             this_insn = insndesc['instruction']
             if this_insn in (name, 'ENV'):
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -86,6 +86,35 @@ USER  {0}""".format(NON_ASCII)
         base_img = dfparser.baseimage
         assert base_img.startswith('fedora')
 
+    def test_get_parent_env(self, tmpdir):
+        tmpdir_path = str(tmpdir.realpath())
+        p_env = {"bar": "baz"}
+        df1 = DockerfileParser(tmpdir_path, env_replace=True, parent_env=p_env)
+        df1.lines = [
+            "FROM parent\n",
+            "ENV foo=\"$bar\"\n",
+            "LABEL label=\"$foo $bar\"\n"
+        ]
+
+        # Even though we inherit an ENV, this .envs count should only be for the
+        # ENVs defined in *this* Dockerfile as we're parsing the Dockerfile and
+        # the parent_env is only to satisfy use of inhereted ENVs.
+        assert len(df1.envs) == 1
+        assert df1.envs.get('foo') == 'baz'
+        assert len(df1.labels) == 1
+        assert df1.labels.get('label') == 'baz baz'
+
+    def test_get_parent_env_from_scratch(self, tmpdir):
+        tmpdir_path = str(tmpdir.realpath())
+        p_env = {"bar": "baz"}
+        df1 = DockerfileParser(tmpdir_path, env_replace=True, parent_env=p_env)
+        df1.lines = [
+            "FROM scratch\n",
+        ]
+
+        assert not df1.envs
+
+
     def test_get_instructions_from_df(self, dfparser, instruction):
         dfparser.content = ""
         lines = []


maxamillion:
https://github.com/DBuildService/dockerfile-parse/issues/20

dockerfile_parse/parser.py
```diff
@@ -91,6 +93,10 @@ def __init__(self, path=None, cache_content=False, env_replace=True):
 
         self.env_replace = env_replace
 
+        if parent_env:
+            logger.debug("Setting inherited parent image ENV vars: %s", parent_env)
+            self.envs(parent_env)
```
twaugh: I was thinking it would be better to change line 293, where `envs = {}`.

We want to know which ENV instructions are declared in the Dockerfile, but when we also want to be able to know correct values for instructions that perform environment variable substitution such as ENV and LABEL.

dockerfile_parse/parser.py
```diff
@@ -66,11 +66,13 @@ def __setitem__(self, key, value):
 
 
 class DockerfileParser(object):
-    def __init__(self, path=None, cache_content=False, env_replace=True):
+    def __init__(self, path=None, cache_content=False, env_replace=True, parent_env={}):
         """
         Initialize path to Dockerfile
         :param path: path to (directory with) Dockerfile
         :param cache_content: cache Dockerfile content inside DockerfileParser
+
```
twaugh: Don't need this blank line.

dockerfile_parse/parser.py
```diff
@@ -66,11 +66,13 @@ def __setitem__(self, key, value):
 
 
 class DockerfileParser(object):
-    def __init__(self, path=None, cache_content=False, env_replace=True):
+    def __init__(self, path=None, cache_content=False, env_replace=True, parent_env={}):
```
lcarva: Please use `parent_env=None`. Using non-primitive types as default values is quite dangerous.

dockerfile_parse/parser.py
lcarva: Wouldn't `self.envs(parent_env)` actually modify the Dockerfile? We definitely don't want that.

maxamillion: @lcarva Isn't this already being used to modify the Dockerfile in place to apply the new labels?

dockerfile_parse/parser.py
twaugh: Caught me. Should say ':param env_replace: bool, whether to perform env var substitution'.

dockerfile_parse/parser.py
```diff
@@ -290,7 +297,7 @@ def _instruction_getter(self, name, env_replace):
         if name != 'LABEL' and name != 'ENV':
             raise ValueError("Unsupported instruction '%s'", name)
         instructions = {}
-        envs = {}
+        envs = self.parent_env
```
lcarva: Do we need to clone `self.parent_env` ? After this method is called it'll contain additional items.

maxamillion: @lcarva I don't follow, what's the issue?

tests/test_parser.py
```diff
@@ -86,6 +86,21 @@ def test_get_baseimg_from_df(self, dfparser):
...
+        df1.lines = [
+            "From parent\n",
+            "ENV foo=\"$bar\"\n",
+            "LABEL label=\"$foo $bar\"\n"
+        ]
+
+        assert len(df1.envs) == 1
+        assert df1.envs.get('foo') == 'baz'
+        assert len(df1.labels) == 1
+        assert df1.labels.get('label') == 'baz baz'
```
twaugh: Can we add another test to the end of this which sets df1.lines to ['FROM scratch\n'] and verifies df1.env only has the variables that came from parent_env? I think that would show up the problem with not using .copy() when assigning envs.

dockerfile_parse/parser.py
maxamillion: @lcarva ohhhh, yes ... should probably `self.parent_env.copy()`

maxamillion:
@twaugh also, I just realized ... given the conditions of the parent image having `{'bar': 'baz'}` shouldn't that be inherited along with the ENV set in the child image mean that the test condition should be `assert len(p.envs) == 2` ?

tests/test_parser.py
maxamillion: Added

dockerfile_parse/parser.py
maxamillion: @lcarva Turns out I was wrong on how the env var was being used here, it's handled correctly in the latest commit

twaugh:
The `envs` attribute needs to be all the environment variables *that were declared in the Dockerfile*. It should disregard any inherited environment variables. That way you can ask questions like "does this Dockerfile declare any environment variables of its own?" easily.

dockerfile_parse/parser.py
```diff
@@ -186,6 +193,18 @@ def _rstrip_backslash(l):
             return l
 
         instructions = []
+        # Handle inherited parent envs
+        for key, val in self.parent_env.items():
+            instructions.append(
```
twaugh: I don't think this is right. The ENV instruction does not appear in the Dockerfile we are parsing. The point of the `structure` attribute is to be able to understand the syntax of the Dockerfile.

The only reason we care about inherited environment variables at all is to be able to know correct values for instructions.

tests/test_parser.py
```diff
@@ -86,6 +86,35 @@ def test_get_baseimg_from_df(self, dfparser):
...
+        tmpdir_path = str(tmpdir.realpath())
+        p_env = {"bar": "baz"}
+        df1 = DockerfileParser(tmpdir_path, env_replace=True, parent_env=p_env)
+        df1.lines = [
+            "FROM parent\n",
+            "ENV foo=\"$bar\"\n",
+            "LABEL label=\"$foo $bar\"\n"
+        ]
+
+        assert len(df1.envs) == 2
```
twaugh: This should be `1`, because there is only 1 ENV instruction in the Dockerfile.

tests/test_parser.py
```diff
@@ -86,6 +86,35 @@ def test_get_baseimg_from_df(self, dfparser):
...
+        assert df1.envs.get('foo') == 'baz'
+        assert df1.envs.get('bar') == 'baz'
```
twaugh: ...and the 'bar' environment variable does not have an ENV instruction in the Dockerfile so should not be in `envs`.

tests/test_parser.py
```diff
@@ -86,6 +86,35 @@ def test_get_baseimg_from_df(self, dfparser):
...
+
+    def test_get_parent_env_from_scratch(self, tmpdir):
+        tmpdir_path = str(tmpdir.realpath())
+        p_env = {"bar": "baz"}
+        df1 = DockerfileParser(tmpdir_path, env_replace=True, parent_env=p_env)
+        df1.lines = [
+            "FROM scratch\n",
+        ]
+
+        assert type(df1.envs.get('bar')) == type("")
```
twaugh: Should be `assert not df1.envs`.

dockerfile_parse/parser.py
maxamillion: Right, but how else can we possibly parse the inherited ENVs if they are not present?

tests/test_parser.py
maxamillion: Wouldn't the inherited ENVs by definition also need to be available to the child image and therefore would be present? Or am I just completely misunderstanding this?

maxamillion: In this comment (https://github.com/DBuildService/dockerfile-parse/pull/21#discussion_r88510197), you request that I add a test to verify that the `parent_env` is inherited and available to the current `p.envs` and now you say that is incorrect. Which is it? :)

twaugh:
If you'd like to squash any commits, please feel free.

maxamillion:
Squashed :)

PR #22, opened 2016-11-21
add type validation to parent_env handling

maxamillion:
Signed-off-by: Adam Miller <maxamillion@fedoraproject.org>

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -92,9 +92,12 @@ class DockerfileParser(object):
 
         self.env_replace = env_replace
 
-        if parent_env:
+        if type(parent_env) is dict:
             logger.debug("Setting inherited parent image ENV vars: %s", parent_env)
             self.parent_env = parent_env
+        elif parent_env != None:
+            logger.debug("Invalid parent image ENV provided: %s", parent_env)
+            self.parent_env = {}
         else:
             self.parent_env = {}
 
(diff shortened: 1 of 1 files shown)


dockerfile_parse/parser.py
```diff
@@ -92,9 +92,12 @@ def __init__(self, path=None, cache_content=False, env_replace=True, parent_env=
 
         self.env_replace = env_replace
 
-        if parent_env:
+        if type(parent_env) is dict:
```
twaugh: Could use `isinstance ()`.

maxamillion: 👍

dockerfile_parse/parser.py
```diff
@@ -92,9 +92,12 @@ def __init__(self, path=None, cache_content=False, env_replace=True, parent_env=
 
         self.env_replace = env_replace
 
-        if parent_env:
+        if isinstance(parent_env, dict):
             logger.debug("Setting inherited parent image ENV vars: %s", parent_env)
             self.parent_env = parent_env
+        elif parent_env != None:
+            logger.debug("Invalid parent image ENV provided: %s", parent_env)
```
lcarva: Don't we want this to raise an exception? It seems unnatural to just bury the problem.

twaugh: If we want to do that, maybe we should just `assert isinstance(parent_env, dict)`.

maxamillion: +1

PR #23, opened 2016-12-09
Add ability to use a file-like object instead of path

lachmanfrantisek:
I have implemented a possibility to cache content with the file-like object. So, it won't be necessary to write `Dockerfile` to disk.

It adds a property `fileobj` to `parser`. If it is not `None`, the content will be written to this `fileobj` instead of the `./Dockerfile` file.

Here is an example of usage with StringIO:

```python
content = r"""
FROM fedora:latest
MAINTAINER Franta

ENV a=aaa
"""

dockerfile = six.StringIO(content)
dfp = dockerfile_parse.DockerfileParser(fileobj=dockerfile,
                                     cache_content=True,
                                     env_replace=False)
```

Can you please have a look at it?

Thanks

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -79,6 +79,10 @@ class DockerfileParser(object):
         :param parent_env: python dict of inherited env vars from parent image
         :param fileobj: file-like object containing Dockerfile content
         """
+
+        if path is not None and fileobj is not None:
+            raise ValueError("Parameters path and fileobj cannot be used together.")
+
         path = path or '.'
         if path.endswith(DOCKERFILE_FILENAME):
             self.dockerfile_path = path
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -77,21 +77,25 @@ class DockerfileParser(object):
         :param path: path to (directory with) Dockerfile
         :param cache_content: cache Dockerfile content inside DockerfileParser
         :param parent_env: python dict of inherited env vars from parent image
-        :param fileobj: file-like object containing Dockerfile content
+        :param fileobj: seekable file-like object containing Dockerfile content
         """
 
-        if path is not None and fileobj is not None:
-            raise ValueError("Parameters path and fileobj cannot be used together.")
+        self.fileobj = fileobj
 
-        path = path or '.'
-        if path.endswith(DOCKERFILE_FILENAME):
-            self.dockerfile_path = path
+        if self.fileobj is not None:
+            if path is not None:
+                raise ValueError("Parameters path and fileobj cannot be used together.")
+            else:
+                self.fileobj.seek(0)
         else:
-            self.dockerfile_path = os.path.join(path, DOCKERFILE_FILENAME)
+            path = path or '.'
+            if path.endswith(DOCKERFILE_FILENAME):
+                self.dockerfile_path = path
+            else:
+                self.dockerfile_path = os.path.join(path, DOCKERFILE_FILENAME)
 
         self.cache_content = cache_content
         self.cached_content = ''  # unicode string
-        self.fileobj = fileobj
 
         if cache_content:
             try:
@@ -114,19 +118,9 @@ class DockerfileParser(object):
     @contextmanager
     def _open_dockerfile(self, mode):
         if self.fileobj is not None:
-            try:
-                self.fileobj.seek(0)
-            except IOError:
-                # not seekable fileobj
-                pass
-
+            self.fileobj.seek(0)
             yield self.fileobj
-
-            try:
-                self.fileobj.seek(0)
-            except IOError:
-                # not seekable fileobj
-                pass
+            self.fileobj.seek(0)
         else:
             with open(self.dockerfile_path, mode) as dockerfile:
                 yield dockerfile
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -119,6 +119,8 @@ class DockerfileParser(object):
     def _open_dockerfile(self, mode):
         if self.fileobj is not None:
             self.fileobj.seek(0)
+            if mode == 'w':
+                self.fileobj.truncate()
             yield self.fileobj
             self.fileobj.seek(0)
         else:
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -73,11 +73,11 @@ class DockerfileParser(object):
                  parent_env=None,
                  fileobj=None):
         """
-        Initialize path to Dockerfile
+        Initialize source of Dockerfile
         :param path: path to (directory with) Dockerfile
         :param cache_content: cache Dockerfile content inside DockerfileParser
         :param parent_env: python dict of inherited env vars from parent image
-        :param fileobj: file-like object used to cache content of Dockerfile
+        :param fileobj: file-like object containing Dockerfile content
         """
         path = path or '.'
         if path.endswith(DOCKERFILE_FILENAME):
@@ -110,8 +110,19 @@ class DockerfileParser(object):
     @contextmanager
     def _open_dockerfile(self, mode):
         if self.fileobj is not None:
-            self.fileobj.seek(0)
+            try:
+                self.fileobj.seek(0)
+            except IOError:
+                # not seekable fileobj
+                pass
+
             yield self.fileobj
+
+            try:
+                self.fileobj.seek(0)
+            except IOError:
+                # not seekable fileobj
+                pass
         else:
             with open(self.dockerfile_path, mode) as dockerfile:
                 yield dockerfile
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -66,12 +66,17 @@ class Envs(dict):
 
 
 class DockerfileParser(object):
-    def __init__(self, path=None, cache_content=False, env_replace=True, parent_env=None):
+    def __init__(self, path=None,
+                 cache_content=False,
+                 env_replace=True,
+                 parent_env=None,
+                 fileobj=None):
         """
         Initialize path to Dockerfile
         :param path: path to (directory with) Dockerfile
         :param cache_content: cache Dockerfile content inside DockerfileParser
         :param parent_env: python dict of inherited env vars from parent image
+        :param fileobj: file-like object used to cache content of Dockerfile
         """
         path = path or '.'
         if path.endswith(DOCKERFILE_FILENAME):
@@ -81,6 +86,7 @@ class DockerfileParser(object):
 
         self.cache_content = cache_content
         self.cached_content = ''  # unicode string
+        self.fileobj = fileobj
 
         if cache_content:
             try:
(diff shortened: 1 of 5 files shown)


twaugh:
Thanks for this.

What's the use-case for doing this? Is it that you don't want changes written to the Dockerfile for some reason?

Is there a reason not to use the `fileobj` parameter for reading the file initially, and so requiring that only one of `filename` and `fileobj` is set?

lcarva:
I think having the ability to use a file like object is fantastic! In previous cases, I had to create temporary folders to get around it.

A couple of concerns:

1. Every time the file-like object is read from or written to, it needs to be re-winded. Otherwise, things get really ugly:

```python
>>> from StringIO import StringIO
>>> s = StringIO('blah')
>>> s.read()
'blah'
>>> s.read()
''
>>> s.write('yo')
>>> s.read()
''
>>> s.seek(0)
>>> s.read()
'blahyo'
>>> s = StringIO('blah')
>>> s.write('foo')
>>> s.read()
'h'
```

2. Instead of repeating the logic to decide whether to use filename or file-like object in multiple places, how about using a context manager that wraps the logic? I can also encapsulate the re-winding logic if needed.

```python
from contextlib import contextmanager

  @contextmanager
   def open_dockerfile(self, mode):
     if self.fileobj is not None:
       self.fileobj.seek(0)
       yield self.fileobj
     else:
       with open(self.dockerfile_path, mode) as dockerfile:
         yield dockerfile

  def lines(self):
     ...
     with open_dockerfile('r') as dockerfile:
       ..
```

lachmanfrantisek:
@twaugh The reason for doing this is, that I want to have only parsing functionality and saving files to disk is unnecessary. I'm only reading data from dockerfile. Also for my testing, it will be better to have everything in memory.

@lcarva Thanks for the tips. It was very useful for me.

> 1. Every time the file-like object is read from or written to, it needs to be re-winded. Otherwise, things get really ugly:

The `seek` explains my problems with `StringIO` in another case. Thanks.

> 2. Instead of repeating the logic to decide whether to use filename or file-like object in multiple places, how about using a context manager that wraps the logic? I can also encapsulate the re-winding logic if needed.

Yes, it would be better to encapsulate it. It looks much nicer with the context manager. I'll edit it and send the next version soon.

lachmanfrantisek:
I've switched it to the context manager. Can someone have a look, please?

But I'm not sure about the seeking. Is it necessary to `seek(0)` also after I wrote to `fileobj`? After writing, I don't leave a `fileobj` with a position at the start.

Do I have to change something in the constructor to test, if there is not also `fileobj` and `path` in the same time?

Thanks

dockerfile_parse/parser.py
```diff
@@ -66,12 +67,17 @@ def __setitem__(self, key, value):
...
+                 parent_env=None,
+                 fileobj=None):
         """
         Initialize path to Dockerfile
         :param path: path to (directory with) Dockerfile
         :param cache_content: cache Dockerfile content inside DockerfileParser
         :param parent_env: python dict of inherited env vars from parent image
+        :param fileobj: file-like object used to cache content of Dockerfile
         """
         path = path or '.'
```
lcarva: Maybe the first thing this method does is raise a `ValueError` if both `path` and `fileobj` are truthy?

dockerfile_parse/parser.py
```diff
@@ -100,6 +107,15 @@ def __init__(self, path=None, cache_content=False, env_replace=True, parent_env=
         else:
             self.parent_env = {}
 
+    @contextmanager
+    def _open_dockerfile(self, mode):
+        if self.fileobj is not None:
+            self.fileobj.seek(0)
+            yield self.fileobj
```
lcarva: It might be worth performing an additional `seek(0)` after this `yield` statement.
This will make it simpler to use this library.

twaugh:
I still don't think I really understand how this should be used.

What would the client calling code look like that would want to use this feature?

dockerfile_parse/parser.py
lachmanfrantisek: Thank you for the review. `ValueError` is good idea. I'll add the condition to the constructor.

dockerfile_parse/parser.py
lachmanfrantisek: Thanks for your advice, I'm not much experienced in this stuff.

lachmanfrantisek:
@twaugh I want to use it in an image builder. It's using dockerpy (without `build` command, only `create`, `commit`, `exec`,..). It manages some extra stuff like layering or secrets during the build.

The builder goes throw the dockerfile and executes every line.

'Only memory dockerfile' is useful for testing this builder and for more general usage of the builder. Also in dockerpy `build`, the dockerfile can  be sent as a `fileobj`. Here is a code snippet from [their older documentation](http://docker-py.readthedocs.io/en/1.10.6/api/):

```python
>>> from io import BytesIO
>>> from docker import Client
>>> dockerfile = '''
... # Shared Volume
... FROM busybox:buildroot-2014.02
... MAINTAINER first last, first.last@yourdomain.com
... VOLUME /data
... CMD ["/bin/sh"]
... '''
>>> f = BytesIO(dockerfile.encode('utf-8'))
>>> cli = Client(base_url='tcp://127.0.0.1:2375')
>>> response = [line for line in cli.build(
...     fileobj=f, rm=True, tag='yourname/volume'
... )]
>>> response
['{"stream":" ---\\u003e a9eb17255234\\n"}',
:
```

Some example from my code:

```python
content = r"""
FROM fedora:latest
MAINTAINER Franta

ENV a=aaa
"""
dockerfile = six.StringIO(content)
dfp = DockerfileParser(fileobj=df, cache_content=True, env_replace=False)

for command in dfp.structure:
    instruction = command["instruction"]
    if instruction == "RUN":
        data_changed = True
        command_id = self.client.exec_create(container=self._current_container_id, cmd=command["value"]).get('Id')
        self.client.exec_start(exec_id=command_id)
        :
```

twaugh:
Oh, so this isn't really about caching but about an alternative to `path`? I like it.

dockerfile_parse/parser.py
```diff
@@ -66,12 +67,17 @@ def __setitem__(self, key, value):
...
 
 class DockerfileParser(object):
-    def __init__(self, path=None, cache_content=False, env_replace=True, parent_env=None):
+    def __init__(self, path=None,
+                 cache_content=False,
+                 env_replace=True,
+                 parent_env=None,
+                 fileobj=None):
         """
         Initialize path to Dockerfile
```
twaugh: Should this be changed? 'Initialize source of Dockerfile' or something?

dockerfile_parse/parser.py
```diff
@@ -66,12 +67,17 @@ def __setitem__(self, key, value):
...
         :param path: path to (directory with) Dockerfile
         :param cache_content: cache Dockerfile content inside DockerfileParser
         :param parent_env: python dict of inherited env vars from parent image
+        :param fileobj: file-like object used to cache content of Dockerfile
```
twaugh: I think this doesn't need to talk about the cache at all. Instead perhaps 'file-like object containing Dockerfile content'?

dockerfile_parse/parser.py
```diff
@@ -100,6 +107,15 @@ def __init__(self, path=None, cache_content=False, env_replace=True, parent_env=
         else:
             self.parent_env = {}
 
+    @contextmanager
+    def _open_dockerfile(self, mode):
+        if self.fileobj is not None:
+            self.fileobj.seek(0)
```
twaugh: Some file-like objects will not support seek() and will raise `IOError`. What should we do in that situation?

lachmanfrantisek: Thanks for mentioning it. Is there any easier way instead of using this?

```python
try:
    self.fileobj.seek(0)
except IOError:
    # not seekable fileobj
    pass
```

lachmanfrantisek:
@twaugh Sorry for confusing description and thanks for tips. I've changed the docs and solved the problem with the not seekable file objects.

@lcarva The second `seek(0)`  after `yield` is there.

dockerfile_parse/parser.py
twaugh: I think that looks right, if that's how we want to handle it.

I just wonder if that's the correct approach. For mode=='r' it seems fine, but for mode=='w' should we fail? We know we've already read from `fileobj` in that case, so we know it's not at the right position.

lachmanfrantisek: I'm not sure. Probably raising an error for `mode=='w'` in second `seek`?

dockerfile_parse/parser.py
```diff
@@ -100,6 +111,26 @@ def __init__(self, path=None, cache_content=False, env_replace=True, parent_env=
         else:
             self.parent_env = {}
 
+    @contextmanager
+    def _open_dockerfile(self, mode):
+        if self.fileobj is not None:
+            try:
+                self.fileobj.seek(0)
+            except IOError:
```
lcarva: Not rewinding the file will cause unexpected behavior when using this class.
Can the "requirement" be that the file must be seekable?

Perhaps this can be enforced with a `seek(0)` call in `__init__` to fail immediately and provide a meaningful error message?

lachmanfrantisek: If there is no real use-case with nonseekable file, than this would be the safest solution.

twaugh: Agreed. Dockerfiles are not so large that they can't be entirely read into an StringIO to be made seekable anyway.

lachmanfrantisek: OK, I will make it this way.

lachmanfrantisek:
Thanks. Can I use module `six` for testing? It would be useful when importing `StringIO`.

twaugh:
Yes, that's a good idea.

lachmanfrantisek:
I'm not sure if we can run same tests for file object as well as for dockerfile from the path.

If there is no problem with that, combining `cache_content` and type of dockerfile in fixtures will probably be the easiest way to test whole functionality or is there another simpler solution?

```diff
-@pytest.fixture(params=[False, True])
+@pytest.fixture(params=[(dockerfile, cache_content) for dockerfile in ['fileobj', 'path'] for cache_content in [True, False]])
 def dfparser(tmpdir, request):
     """
 
     :param tmpdir: already existing fixture defined in pytest
     :param request: parameter, cache_content arg to DockerfileParser
     :return: DockerfileParser instance
     """
-    tmpdir_path = str(tmpdir.realpath())
-    return DockerfileParser(tmpdir_path, request.param)
+
+    if request.param[1] == 'path':
+        tmpdir_path = str(tmpdir.realpath())
+        return DockerfileParser(path=tmpdir_path, cache_content=request.param[0])
+    else:
+        file = six.StringIO()
+        return DockerfileParser(fileobj=file, cache_content=request.param[0])
```

lachmanfrantisek:
During testing I've found that `fileobj` has to be truncated before every write. Isn't it dangerous?

```diff
    def _open_dockerfile(self, mode):
         if self.fileobj is not None:
             self.fileobj.seek(0)
+            if mode == 'w':
+                self.fileobj.truncate()
             yield self.fileobj
             self.fileobj.seek(0)
         else:
```

twaugh:
Let's make sure we note in the constructor docstring that the fileobj will be truncated on write, but yes, good catch.

It's no problem to combine the tests, in fact I like it. :)

tests/fixtures.py
```diff
@@ -10,20 +10,26 @@
 from __future__ import unicode_literals
 
 import pytest
+import six
 
 from dockerfile_parse import DockerfileParser
 
 
-@pytest.fixture(params=[False, True])
+@pytest.fixture(params=[(dockerfile, cache_content) for dockerfile in ['fileobj', 'path'] for cache_content in [True, False]])
```
twaugh: What about if this is `(use_fileobj, cache_content)` so it is also True/False?

tests/fixtures.py
```diff
@@ -10,20 +10,26 @@
...
     """
 
     :param tmpdir: already existing fixture defined in pytest
     :param request: parameter, cache_content arg to DockerfileParser
     :return: DockerfileParser instance
     """
-    tmpdir_path = str(tmpdir.realpath())
-    return DockerfileParser(tmpdir_path, request.param)
+
+    if request.param[1] == 'path':
```
twaugh: We could unpack the tuple as `use_fileobj, cache_content = request.param` to make it a little more readable.

lachmanfrantisek: Thanks for tips. I agree, this looks better.

lachmanfrantisek:
@twaugh , @lcarva Thank both of you for help and advice.

PR #30, opened 2017-01-31
start using tox

rcerven:


twaugh:
Can we add coveralls.io a badge to README.md?

.travis.yml
```diff
@@ -1,16 +1,24 @@
...
+      env: TOXENV=py26
+    - python: "2.7"
+      env: TOXENV=py27
+    - python: "3.4"
+      env: TOXENV=py34
+    - python: "3.5"
+      env: TOXENV=py35
 # command to install dependencies
 install:
   - "pip install -r requirements.txt"
```
vrutkovs: We can remove those lines since Tox now installs the dependencies

twaugh: In theory setup.py would pull in the requirements from requirements.txt, and tox uses setup.py to install the package. This particular setup.py doesn't do that... in fact requirements.txt is empty.

Wonder if we should remove `- "pip install -r requirements.txt"` from this file and remove the requirements.txt file altogether, to make that clearer.

rcerven: well it is empty for now, that might change, and this way it will work in future as well

vrutkovs: >In theory setup.py would pull in the requirements from requirements.txt

This is probably the best option (see https://github.com/vrutkovs/github-review-dashboard/blob/master/setup.py#L15), but I'm not sure if py26 supports that

rcerven: I am using same thing as is in osbs-client and atomic-reactor

setup.py
```diff
@@ -8,9 +8,24 @@
 """
 
 import re
+import sys
 
 from setuptools import setup, find_packages
 
+def _get_requirements(path):
```
vrutkovs: Manual parsing of `requirement.txt` is not a very good idea - it won't work in many cases, e.g.:
```
docker-py < 2.0  #New docker-py has a totally different API
```

setup.py
```diff
@@ -20,5 +35,8 @@
     url='https://github.com/DBuildService/dockerfile-parse',
     license="BSD",
     packages=find_packages(exclude=["tests"]),
+    install_requires=_install_requirements(),
+    setup_requires=['pytest-runner'],
```
lcarva: Keep in mind that this will fail RPM builds, this line must be nuked in spec file. Not sure about `tests_require`.

rcerven: yeah I still need to change spec file

setup.py
rcerven: I tried to use same as in other components, so we can fine tune it later with something better, but as you said your idea doesn't work for 26

setup.py
```diff
@@ -8,9 +8,24 @@
...
+        with open(path) as f:
+            packages = f.read().splitlines()
+    except (IOError, OSError) as ex:
+        raise RuntimeError("Can't open file with requirements: %s", repr(ex))
+    return [p.strip() for p in packages if not re.match(r"^\s*#", p)]
+
+def _install_requirements():
+    requirements = _get_requirements('requirements.txt')
+    if sys.version_info[0] >= 3:
+        requirements += _get_requirements('requirements-py3.txt')
```
lcarva: Looks like `requirements-py3.txt` is an empty file. Place holder for the future?

vrutkovs: I think its a good idea to include this so we could make `setup.py` similar across projects

rcerven: yes, I've created it just to make it same in everywhere, and yeah for the future use

setup.py
twaugh: This change is still needed.

setup.py
vrutkovs: PipSession approach does work for py26

setup.py
vrutkovs: Can we move this to `tests_requires` so we wouldn't have to modify the spec file?

python-dockerfile-parse.spec
```diff
@@ -74,6 +74,8 @@ Python 3 library for Dockerfile manipulation
 %prep
 %setup -n %{project}-%{commit}
 
+# TODO: Fix this properly upstream
+sed -i '/setup_requires/d' setup.py
```
twaugh: Not required now?

rcerven: yop forgot about that when I was reverting it back

Issue #28, opened 2017-01-18
LABEL is not properly handled by dockerfile-parse


phracek:
If Dockerfile is defined like:

~~~
LABEL summary="Postfix is a Mail Transport Agent (MTA)." \
       version = "1.0" \
       description="Postfix is mail transfer agent that routes and delivers mail." \
       io.k8s.description="Postfix is mail transfer agent that routes and delivers mail." \
       io.k8s.display-name="Postfix 3.1" \
       io.openshift.expose-services="10025:postfix" \
io.openshift.tags="postfix,mail,mta"
~~~

then dockerfile-parse parse it like:
~~~
{u'content': u'LABEL summary="Postfix is a Mail Transport Agent (MTA)." \\\n    version="1.0" \\\n    description="Postfix is mail transfer agent that routes and delivers mail." \\\n    io.k8s.description="Postfix is mail transfer agent that routes and delivers mail." \\\n    io.k8s.diplay-name="Postfix 3.1" \\\n    io.openshift.expose-services="10025:postfix" \\\n    io.openshift.tags="postfix,mail,mta"\n',
 u'endline': 22,
 u'instruction': u'LABEL',
 u'startline': 16,
 u'value': u'summary="Postfix is a Mail Transport Agent (MTA)."     version="1.0"     description="Postfix is mail transfer agent that routes and delivers mail."     io.k8s.description="Postfix is mail transfer agent that routes and delivers mail."     io.k8s.diplay-name="Postfix 3.1"     io.openshift.expose-services="10025:postfix"     io.openshift.tags="postfix,mail,mta"'}
~~~

python program as reproducer:
~~~
        dfp = DockerfileParser(path=self.dir)
        import pprint
        for struct in dfp.structure:
            pprint.pprint(struct)
~~~

Would it be possible to somehow structure it?

twaugh:
The `labels` property provides access to individual labels.

lcarva:
Each item in structure is meant to represent an instruction, which can span multiple lines. I'm not sure I understand the request, could you provide example of the behavior you're expecting?

phracek:
Hi folks, I am happy you responded.

If my labels are defined like multiple lines like in my case, I would prefer to create a structure like below.

SEE value. Nowadays it is string, which is wrongly parsable by python.
~~~
[...]

PR #33, opened 2017-04-11
Labels and envs dict to structure

lachmanfrantisek:
As well as @phracek in #28, I need to parse `ENV`/`LABEL` key-values for each instruction.

For a backwards compatibility, I've only added another item to instruction dictionary:
```Dockerfile
FROM fedora:25

LABEL summary="Postfix is a Mail Transport Agent (MTA)." \
       version="1.0" \
       description="Postfix is mail transfer agent that routes and delivers mail." \
       io.k8s.description="Postfix is mail transfer agent that routes and delivers mail." \
       io.k8s.display-name="Postfix 3.1" \
       io.openshift.expose-services="10025:postfix" \
io.openshift.tags="postfix,mail,mta"

```

```python
{
    u'startline': 2,
    u'dictionary': {u'io.k8s.description': u'Postfix is mail transfer agent that routes and delivers mail.',
                    u'description': u'Postfix is mail transfer agent that routes and delivers mail.',
                    u'io.k8s.display-name': u'Postfix 3.1',
                    u'summary': u'Postfix is a Mail Transport Agent (MTA).',
                    u'version': u'1.0',
                    u'io.openshift.expose-services': u'10025:postfix',
                    u'io.openshift.tags': u'postfix,mail,mta'},
    u'instruction': u'LABEL',
    u'value': u'summary="Postfix is a Mail Transport Agent (MTA)."        version="1.0"        description="Postfix is mail transfer agent that routes and delivers mail."        io.k8s.description="Postfix is mail transfer agent that routes and delivers mail."        io.k8s.display-name="Postfix 3.1"        io.openshift.expose-services="10025:postfix" io.openshift.tags="postfix,mail,mta"',
    u'content': u'LABEL summary="Postfix is a Mail Transport Agent (MTA)." \\\n       version="1.0" \\\n       description="Postfix is mail transfer agent that routes and delivers mail." \\\n       io.k8s.description="Postfix is mail transfer agent that routes and delivers mail." \\\n       io.k8s.display-name="Postfix 3.1" \\\n       io.openshift.expose-services="10025:postfix" \\\nio.openshift.tags="postfix,mail,mta"\n',
    u'endline': 8
}
```

There has been restructured/refactored the parsing function and added some basic tests.

What do you all think about it?
Thanks for any suggestions.

-----

I've also removed one test:

```diff
-        lines.insert(-1, '{0} "name1"=\'value 1\' "name2"=myself name3="" name4\n'.format(i))
+        lines.insert(-1, '{0} "name1"=\'value 1\' "name2"=myself name3=""\n'.format(i))

-        assert instructions.get('name4') == ''
```
 I am not sure, but this `ENV`/`LABEL` Dockerfile syntax is not valid:

`LABEL "name1"='value 1' "name2"=myself name3="" name4`

(missing value for `name4`)

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -13,15 +13,18 @@ import json
 import logging
 import os
 import re
+from contextlib import contextmanager
+
+from .constants import DOCKERFILE_FILENAME
+from .util import (b2u, extract_labels_or_envs, get_key_val_dictionary,
+                   remove_quotes, shlex_split, u2b, Context)
+
 try:
     # py3
     from shlex import quote
 except ImportError:
     from pipes import quote
 
-from .constants import DOCKERFILE_FILENAME
-from .util import b2u, u2b, shlex_split, strip_quotes, remove_quotes, remove_nonescaped_quotes, EnvSubst
-from contextlib import contextmanager
 
 logger = logging.getLogger(__name__)
 
@@ -330,40 +333,20 @@ class DockerfileParser(object):
             raise ValueError("Unsupported instruction '%s'", name)
         instructions = {}
         envs = self.parent_env.copy()
-        for insndesc in self.structure:
-            this_insn = insndesc['instruction']
-            if this_insn in (name, 'ENV'):
-                logger.debug("%s value: %r", name.lower(), insndesc['value'])
-                shlex_splits = shlex_split(insndesc['value'],
-                                           env_replace=env_replace, envs=envs)
-                if '=' not in shlex_splits[0]:  # LABEL/ENV name value
-                    # split it to first (name) and the rest (value)
-                    key_val = insndesc['value'].split(None, 1)
-                    key = strip_quotes(key_val[0])
-                    try:
-                        val = key_val[1]
-                    except IndexError:
-                        val = ''
 
-                    if env_replace:
-                        val = EnvSubst(val, envs).substitute()
-
-                    val = remove_nonescaped_quotes(val)
-                    if this_insn == name:
-                        instructions[key] = val
-                        logger.debug("new %s %r=%r", name.lower(), key, val)
-                    if env_replace and this_insn == 'ENV':
-                        envs[key] = val
-                else:  # LABEL/ENV "name"="value"
-                    for token in shlex_splits:
-                        key_val = token.split("=", 1)
-                        key = key_val[0]
-                        val = key_val[1] if len(key_val) > 1 else ''
-                        if this_insn == name:
-                            instructions[key] = val
-                            logger.debug("new %s %r=%r", name.lower(), key, val)
-                        if this_insn == 'ENV':
-                            envs[key] = val
+        for instruction_desc in self.structure:
+            this_instruction = instruction_desc['instruction']
+            if this_instruction in (name, 'ENV'):
+                logger.debug("%s value: %r", name.lower(), instruction_desc['value'])
+                key_val_list = extract_labels_or_envs(env_replace=env_replace,
+                                                      envs=envs,
+                                                      instruction_value=instruction_desc['value'])
+                for key, value in key_val_list:
+                    if this_instruction == name:
+                        instructions[key] = value
+                        logger.debug("new %s %r=%r", name.lower(), key, value)
+                    if this_instruction == 'ENV':
+                        envs[key] = value
 
         logger.debug("instructions: %r", instructions)
         return Labels(instructions, self) if name == 'LABEL' else Envs(instructions, self)
@@ -549,3 +532,25 @@ class DockerfileParser(object):
             lines = self.lines
             lines += new_line
             self.lines = lines
+
+    @property
+    def context_structure(self):
+        """
+        :return: list of Context objects
+            (Contains info about labels and environment variables for each line.)
+        """
+        instructions = []
+        last_context = Context()
+        for instr in self.structure:
+            context = Context(envs=dict(last_context.envs),
+                              labels=dict(last_context.labels))
+            instruction_type = instr['instruction']
+            if instruction_type in ["ENV", "LABEL"]:
+                val = get_key_val_dictionary(instruction_value=instr['value'],
+                                             env_replace=self.env_replace,
+                                             envs=last_context.envs)
+                context.set_line_value(context_type=instruction_type, value=val)
+
+            instructions.append(context)
+            last_context = context
+        return instructions
--- a/dockerfile_parse/util.py
+++ b/dockerfile_parse/util.py
@@ -9,8 +9,8 @@ of the BSD license. See the LICENSE file for details.
 
 from __future__ import print_function, unicode_literals
 
-from io import StringIO
 import shlex
+from io import StringIO
 
 from .constants import PY2
 
@@ -188,3 +188,102 @@ def remove_nonescaped_quotes(string):
     string = remove_quotes(string)
     string = string.replace('\ ', ' ')
     return string.replace("\\s", "'").replace('\\d', '"')  # restore
+
+
+def split_tuple(text):
+    text_split = text.split('=', 1)
+    if len(text_split) == 2:
+        return tuple(text_split)
+    return None
+
+
+def extract_labels_or_envs(env_replace, envs, instruction_value):
+    shlex_splits_raw = shlex_split(instruction_value,
+                                   env_replace=env_replace, envs=envs)
+
+    key_val_list = []
+    if '=' not in shlex_splits_raw[0]:  # LABEL/ENV name value
+        # split it to first (name) and the rest (value)
+        key_val = instruction_value.split(None, 1)
+        key = strip_quotes(key_val[0])
+        try:
+            val = key_val[1]
+        except IndexError:
+            val = ''
+
+        if env_replace:
+            val = EnvSubst(val, envs).substitute()
+        val = remove_nonescaped_quotes(val)
+
+        key_val_list.append((key, val))
+
+    else:  # LABEL/ENV "name"="value"
+
+        for k_v in shlex_splits_raw:
+            key_val_list.append(split_tuple(k_v))
+
+    return key_val_list
+
+
+def get_key_val_dictionary(instruction_value, env_replace=False, envs=None):
+    envs = envs or []
+    return dict(extract_labels_or_envs(instruction_value=instruction_value,
+                                       env_replace=env_replace,
+                                       envs=envs))
+
+
+class Context(object):
+    def __init__(self, envs=None, labels=None, line_envs=None, line_labels=None):
+        """
+        Class representing current state of environment variables and labels.
+
+        :param envs: dict with variables valid for this line
+            (all variables defined to this line)
+        :param labels: dict with labels valid for this line
+            (all labels defined to this line)
+        :param line_envs: dict with variables defined on this line
+        :param line_labels: dict with labels defined on this line
+        """
+        self.envs = envs or {}
+        self.labels = labels or {}
+        self.line_envs = line_envs or {}
+        self.line_labels = line_labels or {}
+
+    def set_line_value(self, context_type, value):
+        """
+        Set value defined on this line ('line_envs'/'line_labels')
+        and update 'envs'/'labels'.
+
+        :param context_type: "ENV" or "LABEL"
+        :param value: new value for this line
+        """
+        if context_type.upper() == "ENV":
+            self.line_envs = value
+            self.envs.update(value)
+        elif context_type.upper() == "LABEL":
+            self.line_labels = value
+            self.labels.update(value)
+
+    def get_line_value(self, context_type):
+        """
+        Get the values defined on this line.
+
+        :param context_type: "ENV" or "LABEL"
+        :return: values of given type defined on this line
+        """
+        if context_type.upper() == "ENV":
+            return self.line_envs
+        elif context_type.upper() == "LABEL":
+            return self.line_labels
+
+    def get_values(self, context_type):
+        """
+        Get the values valid on this line.
+
+        :param context_type: "ENV" or "LABEL"
+        :return: values of given type valid on this line
+        """
+        if context_type.upper() == "ENV":
+            return self.envs
+        elif context_type.upper() == "LABEL":
+            return self.labels
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -16,6 +16,7 @@ import pytest
 import re
 import six
 import sys
+from textwrap import dedent
 
 from dockerfile_parse import DockerfileParser
 from tests.fixtures import dfparser, instruction
@@ -46,9 +47,9 @@ class TestDockerfileParser(object):
         assert setup_py_version == module_version
 
     def test_dockerfileparser(self, dfparser):
-        df_content = """\
-FROM fedora
-CMD {0}""".format(NON_ASCII)
+        df_content = dedent("""\
+            FROM fedora
+            CMD {0}""".format(NON_ASCII))
         df_lines = ["FROM fedora\n", "CMD {0}".format(NON_ASCII)]
 
         dfparser.content = ""
@@ -113,11 +114,11 @@ CMD {0}""".format(NON_ASCII)
                                        'value': 'command2 &&     command3'}]
 
     def test_dockerfile_json(self, dfparser):
-        dfparser.content = """\
-# comment
-From  base
-LABEL foo="bar baz"
-USER  {0}""".format(NON_ASCII)
+        dfparser.content = dedent("""\
+            # comment
+            From  base
+            LABEL foo="bar baz"
+            USER  {0}""").format(NON_ASCII)
         expected = json.dumps([{"FROM": "base"},
                                {"LABEL": "foo=\"bar baz\""},
                                {"USER": "{0}".format(NON_ASCII)}])
@@ -161,7 +162,7 @@ USER  {0}""".format(NON_ASCII)
         dfparser.content = ""
         lines = []
         i = instruction
-        lines.insert(-1, '{0} "name1"=\'value 1\' "name2"=myself name3="" name4\n'.format(i))
+        lines.insert(-1, '{0} "name1"=\'value 1\' "name2"=myself name3=""\n'.format(i))
         lines.insert(-1, '{0} name5=5\n'.format(i))
         lines.insert(-1, '{0} "name6"=6\n'.format(i))
         lines.insert(-1, '{0} name7\n'.format(i))
@@ -171,6 +172,7 @@ USER  {0}""".format(NON_ASCII)
         lines.insert(-1, '{0} "name1 1"=1\n'.format(i))
         lines.insert(-1, '{0} "name12"=12 \ \n   "name13"=13\n'.format(i))
         lines.insert(-1, '{0} name14=1\ 4\n'.format(i))
+        lines.insert(-1, '{0} name15="with = in value"\n'.format(i))
         # old syntax (without =)
         lines.insert(-1, '{0} name101 101\n'.format(i))
         lines.insert(-1, '{0} name102 1 02\n'.format(i))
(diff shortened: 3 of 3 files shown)


twaugh:
Could you please elaborate on the use-case for this?

lachmanfrantisek:
Thank you for your response.

I'm working on a [builder for docker images](https://gitlab.com/lachmanfrantisek/incubator) and need to take labels/variables per instruction. (Not for the whole dockerfile.)

Example:

```dockerfile
FROM fedora:25
ENV key=value
RUN command $key
ENV key=another_value key2=value2
```

When calling `RUN command $key`, I need only variables defined in the previous lines. (The `envs` property of parser is not valid in that case.)

lachmanfrantisek:
Or what about having separate properties `label_structure` and `env_strucutre`, with valid state for each instruction? Example would be probably better to understand:

```dockerfile
FROM fedora:25
ENV key=value
RUN command $key
LABEL label1="description 1" label2="description2"
ENV key=another_value key2=value2
```

```python
dockerfile_parser.env_structure = [
{},
{'key': 'value'},
{'key': 'value'},
{'key': 'value'},
{'key': 'another_value', 'key2': 'value2'}
]
```

```python
dockerfile_parser.label_structure = [
{},
{},
{},
{'label1': 'description 1', 'label2': 'description2'},
{'label1': 'description 1', 'label2': 'description2'}
]
```

This is precisely what I need, but is this more usefull for others than the previous aproach?

lcarva:
I'm wondering if having a way to iterate over each instructions while accessing the current context is a more intuitive approach. Something along the lines of:

```python
for instruction, context in self.interactive_instructions:
  print(context.label)  # {'label1': 'description 1'}
  print(context.env)  # {'env1': 'value 1'}
  # instruction is the same as the items in self.structure
```

The `context` would obviously be read-only.

lachmanfrantisek:
@lcarva, iterating over each instruction would be a good idea. Thank you.

I'm just thinking about a value of `context.label`/`context.env`.

What about this example?

```dockerfile
FROM fedora:25
ENV key=value
ENV key2=$key
```

There are three options of `context.env` for last line:
1. `{'key2': '$key'}` - raw values for this line
2. `{'key2': 'value'}` - value for this line with variable expansion
3. `{'key': 'value', 'key2': 'value'}` - all variables valid on this line

The first is probably useless, but second and third would be useful. Provide both of them can be also a solution.

twaugh:
I think that during iteration, the `env_replace` parameter provided at construction time should be honoured. If some caller wants to see both the raw and substituted values at the same time they should `zip` two iterators together.

lachmanfrantisek:
Thank you for your tips.

Updated version should provide following results:

`Dockerfile`:
```dockerfile
FROM fedora:25
LABEL a=b
ENV key=value
ENV key2=$key
```

Code sample:
```python
for context in dfp.context_structure:
    print("envs: {}\n"
          "labels: {}\n"
          "line_envs: {}\n"
          "line_labels: {}\n"
          .format(context.envs, context.labels,
                  context.line_envs, context.line_labels))
```

`env_replace=True`:
```python
envs: {}
labels: {}
line_envs: {}
line_labels: {}

envs: {}
labels: {u'a': u'b'}
line_envs: {}
line_labels: {u'a': u'b'}

envs: {u'key': u'value'}
labels: {u'a': u'b'}
line_envs: {u'key': u'value'}
line_labels: {}

envs: {u'key2': u'value', u'key': u'value'}
labels: {u'a': u'b'}
line_envs: {u'key2': u'value'}
line_labels: {}
```


`env_replace=False`:
```python
envs: {}
labels: {}
line_envs: {}
line_labels: {}

envs: {}
labels: {u'a': u'b'}
line_envs: {}
line_labels: {u'a': u'b'}

envs: {u'key': u'value'}
labels: {u'a': u'b'}
line_envs: {u'key': u'value'}
line_labels: {}

envs: {u'key2': u'$key', u'key': u'value'}
labels: {u'a': u'b'}
line_envs: {u'key2': u'$key'}
line_labels: {}
```

What do you think about this?

TomasTomecek:
I have a question to behaviour:

Here is my program:
```python

dfp = DockerfileParser()

dfp.content = """FROM x
LABEL x=y
RUN x
LABEL a=b c=d
ENV q=1
 w=2
LABEL i=8"""

for context in dfp.context_structure:
    print("envs: {}\n"
          "labels: {}\n"
          "line_envs: {}\n"
          "line_labels: {}\n"
          .format(context.envs, context.labels,
                  context.line_envs, context.line_labels))
```

And here's the output:
```
envs: {}
labels: {}
line_envs: {}
line_labels: {}

envs: {}
labels: {u'x': u'y'}
line_envs: {}
line_labels: {u'x': u'y'}

envs: {}
labels: {}
line_envs: {}
line_labels: {}

envs: {}
labels: {u'a': u'b', u'x': u'y', u'c': u'd'}
line_envs: {}
line_labels: {u'a': u'b', u'c': u'd'}

envs: {u'q': u'1'}
labels: {u'a': u'b', u'x': u'y', u'c': u'd'}
line_envs: {u'q': u'1'}
line_labels: {}

envs: {u'q': u'1'}
labels: {u'a': u'b', u'x': u'y', u'c': u'd', u'i': u'8'}
line_envs: {}
line_labels: {u'i': u'8'}
```

I would expect that the third item would look like this:
```
envs: {}
labels: {u'x': u'y'}
line_envs: {}
line_labels: {}
```

Is this bug or do I just don't understand it?

lachmanfrantisek:
@TomasTomecek thank you for finding the bug. It should be corrected and covered by some tests.

Is the documentation OK?

TomasTomecek:
LGTM now, thank you!

Luiz, Tim: waiting for your approval, don't want to merge this myself. I believe @lachmanfrantisek would also appreciate cutting a new release afterwards. I can help with builds in Fedora.

dockerfile_parse/util.py
```diff
@@ -188,3 +188,105 @@ def remove_nonescaped_quotes(string):
...
+        elif context_type.upper() == "LABEL":
+            self.line_labels = value
+
+    def get_line_value(self, context_type):
+        """
+        Get the values defined on this line.
+
+        :param context_type: "ENV" or "LABEL"
+        :return: values of given type defined on this line
+        """
```
twaugh: I wonder if this is needed. Aren't self.line_envs and self.line_labels (and for get_values(), self.envs and self.labels) enough?

Similarly with update_line_value(), why have that but no update_value()?

dockerfile_parse/parser.py
```diff
@@ -549,3 +532,28 @@ def _add_instruction(self, instruction, value):
...
+
+    @property
+    def context_structure(self):
+        """
+        :return: list of context objects
+            (Contains info about labels and environment variables for each line.)
+        """
+        instructions = []
+        envs = {}
+        labels = {}
```
twaugh: Why track envs and labels separately from what Context() stores?

dockerfile_parse/util.py
lachmanfrantisek: > I wonder if this is needed. Aren't self.line_envs and self.line_labels (and for get_values(), self.envs and self.labels) enough?

It is not so necessary, but useful when accessing the properties by name. (It is also handy for testing.)

> Similarly with update_line_value(), why have that but no update_value()?

The `update_line_value()` is only a shortcut. It looks cleaner to me. Shall I add also an `update_value()` or remove them both?

dockerfile_parse/parser.py
lachmanfrantisek: Thank you for your response. Is this better?

```python
@property
    def context_structure(self):
        """
        :return: list of context objects
            (Contains info about labels and environment variables for each line.)
        """
        instructions = []
        last_context = Context()
        for instr in self.structure:
            context = Context(envs=dict(last_context.envs),
                              labels=dict(last_context.labels))
            instruction_type = instr['instruction']
            if instruction_type in ["ENV", "LABEL"]:
                val = get_key_val_dictionary(instruction_value=instr['value'],
                                             env_replace=self.env_replace,
                                             envs=last_context.envs)
                context.update_line_value(context_type=instruction_type, value=val)
                context.envs.update(context.line_envs)
                context.labels.update(context.line_labels)

            instructions.append(context)
            last_context = context
        return instructions
```

And what about updating of `context.envs`/`context.labels` in `context.update_line_value`? So we can remove these lines:
```python
context.envs.update(context.line_envs)
context.labels.update(context.line_labels)
```

twaugh: Yes, I like this better,  and having `context.update_line_value` behave that way makes sense to me.

lachmanfrantisek:
@twaugh, the `context_structure` has been updated. Is it OK?

TomasTomecek:
I really like the last commit, looks a lot more cleaner.

dockerfile_parse/parser.py
```diff
@@ -549,3 +532,25 @@ def _add_instruction(self, instruction, value):
             lines = self.lines
             lines += new_line
             self.lines = lines
+
+    @property
+    def context_structure(self):
+        """
+        :return: list of context objects
```
twaugh: How about "list of Context objects" (use actual object type name)?

lachmanfrantisek: Thank you. Fixed.

dockerfile_parse/util.py
```diff
@@ -188,3 +188,108 @@ def remove_nonescaped_quotes(string):
...
+            val = ''
+
+        if env_replace:
+            val = EnvSubst(val, envs).substitute()
+        val = remove_nonescaped_quotes(val)
+
+        key_val_list.append((key, val))
+
+    else:  # LABEL/ENV "name"="value"
+        shlex_splits = []
```
lcarva: iiuc, `shlex_splits` is a list of strings in the form of `'key=value'`. It seems that this can be better handled by using a different data structure, either a dict or a list of tuples. Something like:

```python
shlex_splits = {}
current_key = None
for part in shlex_splits_raw:
  if current_key:
    shlex_splits[current_key] = part
    current_key = None
    continue
  if part != '=':
    current_key = part

key_val_list.extend(shlex_splits.items())
```

dockerfile_parse/parser.py
```diff
@@ -330,40 +333,20 @@ def _instruction_getter(self, name, env_replace):
...
-                        val = ''
-
-                    if env_replace:
-                        val = EnvSubst(val, envs).substitute()
-
-                    val = remove_nonescaped_quotes(val)
-                    if this_insn == name:
-                        instructions[key] = val
-                        logger.debug("new %s %r=%r", name.lower(), key, val)
-                    if env_replace and this_insn == 'ENV':
```
lcarva: The `env_replace` check seems to have been removed. This is probably fine, especially since this was only being done for this instruction format, and not when `"="` was used.

dockerfile_parse/util.py
lcarva: I agree that the getters seem overkill. Using `context.envs` versus `context.get_values('ENV')` is a much better interface IMO.

I do see the benefit of `set_line_value`, but am wondering if this would be better handled by passing the instruction to `Context.__init__` for better encapsulation. Either way, I'm also fine with the existing setter approach.

tests/test_parser.py
```diff
@@ -500,3 +499,172 @@ def test_path_and_fileobj_together(self):
     def test_nonseekable_fileobj(self):
         with pytest.raises(AttributeError):
             DockerfileParser(fileobj=sys.stdin)
+
+    @pytest.mark.parametrize('instruction', [
+        'LABEL',
+        'ENV'
+    ])
+    def test_context_structure_per_line(self, dfparser, instruction):
+        dfparser.content = """
```
lcarva: FYI, consider using `textwrap.dedent` so you can indent this string:

```python
from textwrap import dedent
def foo():
   s = dedent("""\
       multi
       line
       string!!""")

  # Which is the same as:
  s = """
multi
line
string!!"""
```

lcarva:
This is looking really good! Just a few comments regarding code style.

dockerfile_parse/util.py
lachmanfrantisek: Thanks for a suggestion.

Unfortunately, in `shlex_splits_raw`, there can be `["key", "=", "value]` as well as `["key=value"]`. I have only changed the second `if`.

dockerfile_parse/parser.py
lachmanfrantisek: What about replacing the original value from instruction?

```diff
 def extract_labels_or_envs(env_replace, envs, instruction_value):
+    if env_replace:
+        instruction_value = EnvSubst(instruction_value, envs).substitute()
+
    shlex_splits_raw = shlex_split(instruction_value,
                                   env_replace=env_replace, envs=envs)

    key_val_list = []
    if '=' not in shlex_splits_raw[0]:  # LABEL/ENV name value
        # split it to first (name) and the rest (value)
        key_val = instruction_value.split(None, 1)
        key = strip_quotes(key_val[0])
        try:
            val = key_val[1]
        except IndexError:
            val = ''
 
-        if env_replace:
-            val = EnvSubst(val, envs).substitute()
         val = remove_nonescaped_quotes(val)
 ```

File: tests/test_parser.py (hunk 6 above)
Comment (lachmanfrantisek) [inline tests/test_parser.py]: Thank you, don't know it.

I only need to add extra space after backslash in `{0} description="This text illustrates \ ` here:

```python
        dfparser.content = (dedent("""
            FROM fedora:25
:
:
            {0} description="This text illustrates \ 
            that label-values can span multiple lines."
            """)).format(instruction)
```

TomasTomecek:
@lachmanfrantisek I think that github doesn't notify when you add/change commits so you should add a comment whenever you add something :)

@lcarva ready for another review round?

lachmanfrantisek:
@TomasTomecek thanks, good to know..;-)

lachmanfrantisek:
@lcarva thank you. And what do you think of the `env_replace`? Should I have to change it to this?

```diff
 def extract_labels_or_envs(env_replace, envs, instruction_value):
+    if env_replace:
+        instruction_value = EnvSubst(instruction_value, envs).substitute()
+
    shlex_splits_raw = shlex_split(instruction_value,
                                   env_replace=env_replace, envs=envs)

    key_val_list = []
    if '=' not in shlex_splits_raw[0]:  # LABEL/ENV name value
        # split it to first (name) and the rest (value)
        key_val = instruction_value.split(None, 1)
        key = strip_quotes(key_val[0])
        try:
            val = key_val[1]
        except IndexError:
            val = ''
 
-        if env_replace:
-            val = EnvSubst(val, envs).substitute()
         val = remove_nonescaped_quotes(val)
 ```

**Comment by TomasTomecek** [comment]
@lachmanfrantisek we need to match behaviour of docker here, can you check how docker behaves? I guess that something like this could be a good test case:

```
ENV x="a b 'c'"
ENV y="'d' e 'f g'"
LABEL n=$x m="$y" o='$x $y'

lcarva:
> And what do you think of the env_replace? Should I have to change it to this?

Can you elaborate on what that means? I'm not sure I fully understand the proposed change.

lachmanfrantisek:
@TomasTomecek yes, docker parses your example as follows:

- envs:
```
"x=a b 'c'",
"y='d' e 'f g'"
```

- labels:
```
"m": "'d' e 'f g'",
"n": "a b 'c'",
"o": "$x $y"
```

@lcarva my proposed change does not work on the previous example by @TomasTomecek . The current version of dockerfile-parse is failing as well. I'm working on it for some hours and still not finished. It is not so easy as it looks. It needs huge changes on parsing and escaping quotes.

What about moving it to another issue?

lcarva:
> What about moving it to another issue?

+1 on that! Let's isolate these changes.

lachmanfrantisek:
> +1 on that! Let's isolate these changes.

Ok, I can look at it later.

lachmanfrantisek:
@lcarva I've created an issue #34 for the problem with parsing.

Can we merge this as it is?

lcarva:
> Can we merge this as it is?

I think so. @twaugh wanna take a look first?

TomasTomecek:
I'm also +1 on merging this if the behaviour with quotes is the same -- I would say it's kind of a corner case, usually the env vars and labels are "sane".

lachmanfrantisek:
@twaugh @lcarva Are there any problems with the merging?

Thanks.

TomasTomecek:
I checked with @twaugh today and there is only one problem: time.

lachmanfrantisek:
OK, I'll wait. No problem.

dockerfile_parse/util.py
```diff
@@ -188,3 +188,113 @@ def remove_nonescaped_quotes(string):
     string = remove_quotes(string)
     string = string.replace('\ ', ' ')
     return string.replace("\\s", "'").replace('\\d', '"')  # restore
+
+
+def split_tuple(text):
+    text_split = text.split('=')
```
twaugh: What about if a value contains '=', e.g. foo=bar=baz? Should this be `split('=', 1)` (like the previous code) or is it intentional the way it is?

dockerfile_parse/util.py
```diff
@@ -188,3 +188,113 @@ def remove_nonescaped_quotes(string):
...
+        val = remove_nonescaped_quotes(val)
+
+        key_val_list.append((key, val))
+
+    else:  # LABEL/ENV "name"="value"
+        shlex_splits = {}
+        current_key = None
+        for part in shlex_splits_raw:
+            if current_key:
+                shlex_splits[current_key] =     part
```
twaugh: What's happened with whitespace here?

dockerfile_parse/util.py
lachmanfrantisek: Thanks, good point. I'll correct it.

lachmanfrantisek:
@twaugh both points should be corrected. Thanks for such a deep review and all suggestions.

dockerfile_parse/util.py
twaugh: Thanks. Would you mind also adding a test case for that? It's a shame it wasn't caught by unit tests.

dockerfile_parse/parser.py
twaugh: I'm trying to work out why `EnvSubst` is only used in the "name value" case, and not in the "name=value" case, and why the tests pass like that. Any idea?

By the way I don't think we can call it prior to stripping any outer quotes.

lachmanfrantisek: The old version has it like that. The another substitution is done in this:

```python
    shlex_splits_raw = shlex_split(instruction_value,
                                   env_replace=env_replace, envs=envs)
```

lachmanfrantisek: The all parsing part probably needs a deeper look, we can look at it in the #34?

lachmanfrantisek:
@twaugh I've added some tests for the `key = "with = in value"` format.

dockerfile_parse/util.py
```diff
@@ -188,3 +188,113 @@ def remove_nonescaped_quotes(string):
...
+        for part in shlex_splits_raw:
+            if current_key:
+                shlex_splits[current_key] = part
+                current_key = None
+            elif part != '=':
+                k_v = split_tuple(part)
+                if k_v:
+                    shlex_splits.update([k_v])
+                else:
+                    current_key = part
```
twaugh: I don't follow the logic here (L221-232), and it's quite different to the code it replaces. What's the intention?

`shlex_splits_raw` is a list of name=value strings at L221. Why are we combining them?

dockerfile_parse/parser.py
twaugh: Sure, just trying not to introduce a change in behaviour we don't intend. Thanks for pointing out that shlex_split() performs its own substitution, I'd forgotten.

dockerfile_parse/util.py
lachmanfrantisek: The `shlex_splits_raw` can be also `["key", "=", "value"]`. I'm not sure why, but there were some of such a values..

twaugh: Really? What input generates that? It doesn't sound right.

lachmanfrantisek: Sorry, it was my fault. I can't find it now. It was probably some invalid value.
Now it should be correct/simplified.

twaugh:
Please squash commits into logical changes (perhaps only 1 commit?); then I'll rebase and merge.

lachmanfrantisek:
@twaugh is it OK?

Thank you very much for all your reviews and advice.

twaugh:
Thanks very much for being patient. :)

lachmanfrantisek:
@twaugh As @TomasTomecek said, I would greatly appreciate the release of the new version. Is that possible, please?

Thanks in advance.

twaugh:
I think the spec file needs sorting out first. There were some changes made in the Fedora spec file that I've been trying to port, most of them make sense (but some are wrong).

lachmanfrantisek:
OK, maybe @TomasTomecek can help with this?

twaugh:
Actually now I look more closely it seems I already merged that. :)

lachmanfrantisek:
Thanks..;-)

PR #40, opened 2017-10-04
Fixes #39 Fix cache_content handling

### LINKED ISSUE #39: DockerfileParse does not handle cache_content properly
In our project meta-test-family (https://github.com/fedora-modularity/meta-test-family), used for testing containers, we parse Dockerfile and sometimes rendered, named e.g. Dockerfile.rendered.

Unfortunately there is a bug, from my point of view:

~~~

# dockerfile refers to Dockerfile.rendered and Dockerfile itself looks like https://github.com/container-images/memcached/blob/master/Dockerfile
with open(dockerfile, 'r') as f:
       cached_content = f.readlines()
dfp = DockerfileParser(path=os.path.dirname(dockerfile), cache_content=cached_content)
for struct in dfp.structure:
    print(struct)
~~~

This prints Dockerfile and not Dockerfile.rendered.
I guess, bug is here https://github.com/DBuildService/dockerfile-parse/blob/master/dockerfile_parse/parser.py#L176
Before with there should be
~~~
if cache_content:
    self.cached_content = cache_content
else:
    with ....
~~~

I will sent PR for it.

phracek:
Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>

This Pull Request fixes parsing Dockerfile in case it is used cache_content argument
during initialization of DockerfileParser class.

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -173,11 +173,15 @@ class DockerfileParser(object):
             return self.cached_content
 
         try:
-            with self._open_dockerfile('r') as dockerfile:
-                content = b2u(dockerfile.read())
-                if self.cache_content:
-                    self.cached_content = content
-                return content
+            if self.cache_content:
+                self.cached_content = ''.join(self.cache_content)
+                return self.cached_content
+            else:
+                with self._open_dockerfile('r') as dockerfile:
+                    content = b2u(dockerfile.read())
+                    if self.cache_content:
+                        self.cached_content = content
+                    return content
         except (IOError, OSError) as ex:
             logger.error("Couldn't retrieve content of dockerfile: %r", ex)
             raise


dockerfile_parse/parser.py
```diff
@@ -173,11 +173,15 @@ def content(self):
...
-                if self.cache_content:
-                    self.cached_content = content
-                return content
+            if self.cache_content:
+                self.cached_content = ''.join(self.cache_content)
+                return self.cached_content
+            else:
+                with self._open_dockerfile('r') as dockerfile:
+                    content = b2u(dockerfile.read())
+                    if self.cache_content:
```
lcarva: This is an unreachable if-statement.

dockerfile_parse/parser.py
```diff
@@ -173,11 +173,15 @@ def content(self):
             return self.cached_content
```
lcarva: Should the change be here?
```python
if self.cache_content:
  if not self.cached_content:
    self.cached_content = ''.join(self.cache_content)
  return self.cached_content
```

phracek:
Problem is fixed by the newest upstream version in fedora.

PR #41, opened 2017-10-13
Do not overwrite multiple FROMs for multi-stage builds

hubx:
Hi,

Since 17.05 multiple FROMs are possible see
https://docs.docker.com/engine/reference/builder/#from

Maybe we want to make modifying all baseimage possible at alter stage, for now don't corrupt parsed Dockerfiles

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -228,6 +228,27 @@ class TestDockerfileParser(object):
         dfparser.cmd = CMD[1]
         assert dfparser.cmd == CMD[1]
 
+    def test_modify_from_multistage(self, dfparser):
+        FROM = ('base:${CODE_VERSION}', 'extras:${CODE_VERSION}')
+        CMD = ('/code/run-app', '/code/run-extras')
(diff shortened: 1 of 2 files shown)


twaugh:
Could you please re-base against current master?

tests/test_parser.py
```diff
@@ -228,6 +228,27 @@ def test_modify_instruction(self, dfparser):
...
+CMD  {1}
+
+FROM {2}
+CMD  {3}""").format(FROM[0], CMD[0], FROM[1], CMD[1])
+
+        dfparser.content = df_content
+
+        assert dfparser.baseimage == FROM[0]
+        dfparser.baseimage = FROM[1]
+        assert dfparser.baseimage == FROM[1]
```
rcerven: this doesn't actually test that the second 'from' wasn't modified, as getter is returning 1st, but we aren't checking value of 2nd 'from'

lcarva:
Commit pulled as part of https://github.com/DBuildService/dockerfile-parse/pull/48

PR #42, opened 2017-11-20
Run CI tests in containers

twaugh:


.travis.yml
```diff
@@ -1,22 +1,36 @@
...
+      OS_VERSION=25
+      PYTHON_VERSION=2
+    - OS=fedora
+      OS_VERSION=25
+      PYTHON_VERSION=3
+    - OS=fedora
+      OS_VERSION=26
+      PYTHON_VERSION=2
+    - OS=fedora
+      OS_VERSION=26
```
TomasTomecek: Since 27 is released, I think it would make sense to add it to the list.

TomasTomecek:
Travis enables the use of cronjobs. Do you think it would make sense to run rawhide tests nightly?

.travis.yml
vrutkovs: Rawhide is sufficiently close to 27 now, I was thinking it makes sense to wait until F25 goes EOL and replace it with 27.

Otherwise we'll have an increased test length and probably some 27-specific hacks

TomasTomecek: ...and we'll be sure the software works on latest platforms. Isn't that the reason to use CI?

+1 to do s/25/27/

anyway, I don't want my comment to block this PR

twaugh:
Yes, I'd like us to switch F25 -> F27 across all the OSBS projects. So this PR is to sync up to osbs-client+atomic-reactor etc, and separately we'll switch to F27.

Issue #44, opened 2017-12-19
Concatenated variables not expanded correctly


twaugh:
Given this Dockerfile:
```
FROM scratch
ENV NAME=name VER=1
LABEL component="$NAME$VER"
```

the `component` label should be set as "name1" but it is not. Test case:
```
from dockerfile_parse import DockerfileParser
from io import BytesIO
dockerfile = BytesIO(b'''\
FROM scratch
ENV NAME=name VER=1
LABEL component="$NAME$VER"
''')
p = DockerfileParser(fileobj=dockerfile)
assert p.labels['component'] == 'name1'
```

Instead, `p.labels['component']` is `"name$VER"`. Only the first environment variable was expanded.

PR #45, opened 2017-12-19
Correctly expand concatenated variables

pkubatrh:
Moved the environment variable substitution into a while loop that runs until the next character is not '$'.

Fixes: #44

### DIFF
--- a/dockerfile_parse/util.py
+++ b/dockerfile_parse/util.py
@@ -95,31 +95,36 @@ class EnvSubst(object):
                 continue
 
             if ch == '$' and self.quotes != self.SQUOTE:
-                # Substitute environment variable
-                braced = False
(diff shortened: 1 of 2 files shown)


lcarva:
Thanks for this!

Looks like the travis build for centos failed which seems to be unrelated to your changes. I re-triggered it, but we might need to fix something there. Stay tuned.

Can we add a unit test for these changes?

pkubatrh:
>Can we add a unit test for these changes?

Added a simple test case from #44

lcarva:
@pkubatrh, we fixed the travis errors in https://github.com/DBuildService/dockerfile-parse/pull/46, could you please rebase your branch?

pkubatrh:
Rebased against latest master

PR #46, opened 2017-12-19
travis fixes

lcarva:


twaugh:
I think it needs the pin-version fix. On it.

lcarva:
> I think it needs the pin-version fix.

Yup. Should be all set now.

PR #47, opened 2018-01-10
Tests/disable rawhide

lcarva:


lcarva:
@bfontecc007 @rcerven PTAL

twaugh:
Not updating release notes as this only relates to testing.

PR #48, opened 2018-01-12
Do not overwrite multiple FROMs for multi-stage builds

lcarva:
Since 17.05 multiple FROMs are possible see
https://docs.docker.com/engine/reference/builder/#from

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -494,6 +494,8 @@ class DockerfileParser(object):
                 del lines[insn['startline']:insn['endline'] + 1]
                 lines.insert(insn['startline'], new_line)
                 self.lines = lines
+                if insn['instruction'] == 'FROM':  # Only overwrite first base-image
+                    return
 
     def _delete_instructions(self, instruction, value=None):
         """
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -494,7 +494,7 @@ class DockerfileParser(object):
                 del lines[insn['startline']:insn['endline'] + 1]
                 lines.insert(insn['startline'], new_line)
                 self.lines = lines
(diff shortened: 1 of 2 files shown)


dockerfile_parse/parser.py
```diff
@@ -494,6 +494,8 @@ def _modify_instruction(self, instruction, new_value):
                 del lines[insn['startline']:insn['endline'] + 1]
                 lines.insert(insn['startline'], new_line)
                 self.lines = lines
+                if insn['instruction'] == 'FROM':  # Only overwrite first base-image
```
twaugh: Could be `if instruction == 'FROM':`, and that might make the intent clearer too.

tests/test_parser.py
```diff
@@ -236,6 +236,27 @@ def test_modify_instruction(self, dfparser):
...
+CMD  {1}
+
+FROM {2}
+CMD  {3}""").format(FROM[0], CMD[0], FROM[1], CMD[1])
+
+        dfparser.content = df_content
+
+        assert dfparser.baseimage == FROM[0]
+        dfparser.baseimage = FROM[1]
+        assert dfparser.baseimage == FROM[1]
```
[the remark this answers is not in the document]
lcarva: Great catch! Fixed.

dockerfile_parse/parser.py
lcarva: Fixed.

lcarva:
@twaugh, @rcerven: comments have been addressed. I also did some additional enhancements to `test_parser.py`. Let me know what you think (on Monday, that is 😉 ).

twaugh:
Holding off merging until the next release.

lcarva:
Updated release notes draft:
> Do not overwrite multiple FROMs for multi-stage builds

PR #50, opened 2018-01-17
delete duplicate instruction as it is fixture

rcerven:


lcarva:
Not updating release notes draft because this is a change solely to unit tests.

PR #51, opened 2018-01-17
fix parsing multiline unquoted labels/envs

### LINKED ISSUE #28: LABEL is not properly handled by dockerfile-parse

TomasTomecek:
Fixes #28

All tests were parsing, I mean passing, for me locally.

### DIFF
--- a/dockerfile_parse/util.py
+++ b/dockerfile_parse/util.py
@@ -225,9 +225,15 @@ def extract_labels_or_envs(env_replace, envs, instruction_value):
         key_val_list.append((key, val))
 
     else:  # LABEL/ENV "name"="value"
-
+        last_item = ""
         for k_v in shlex_splits_raw:
-            key_val_list.append(split_tuple(k_v))
+            if '=' not in k_v:
+                last_item += " " + k_v
+            else:
+                if last_item:
+                    key_val_list.append(split_tuple(last_item))
+                last_item = k_v
+        key_val_list.append(split_tuple(last_item))
 
     return key_val_list
 
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -694,3 +694,41 @@ class TestDockerfileParser(object):
             LABEL component="$NAME$VER"
         """)
         assert dfparser.labels['component'] == 'name1'
+
+    def test_multiline(self, dfparser):
+        dfparser.content = dedent("""
+            LABEL a=b c d e \
+                  b="parsing is so hard" \
+                  c=y e s
+        """)
+        assert dfparser.labels == {
+            "a": "b c d e",
+            "b": "parsing is so hard",
+            "c": "y e s"
+        }
+
+    def test_multiline2(self, dfparser):
+        dfparser.content = dedent("""
+            LABEL a="b c" \
+                  b=parsing is so hard \
+                  c="y e s"
+        """)
+        assert dfparser.labels == {
+            "a": "b c",
+            "b": "parsing is so hard",
+            "c": "y e s"
+        }
+
+    def test_multispace_label(self, dfparser):
+        # this fails with docker but succeeds with us b/c shlex.split eats consecutive spaces
+        dfparser.content = "LABEL a=b  c"
+        assert dfparser.labels == {"a": "b c"}
+
+    def test_broken_multiline_label(self, dfparser):
+        # once again, shlex.split is smarter than dockerd's parser
+        dfparser.content = dedent("""
+            LABEL a=b \
+                  c
+        """)
+        assert dfparser.labels == {"a": "b c"}
+


dockerfile_parse/util.py
```diff
@@ -225,9 +225,15 @@ def extract_labels_or_envs(env_replace, envs, instruction_value):
         key_val_list.append((key, val))
 
     else:  # LABEL/ENV "name"="value"
-
+        last_item = None
```
lcarva: Should this be an empty str?

tests/test_parser.py
```diff
@@ -694,3 +694,15 @@ def test_expand_concatenated_variables(self, dfparser):
             LABEL component="$NAME$VER"
         """)
         assert dfparser.labels['component'] == 'name1'
+
+    def test_multiline(self, dfparser):
+        dfparser.content = dedent("""
+            LABEL a=b c d e \
```
lcarva: What happens if there are multiple spaces between characters?

TomasTomecek: oh my, someone would actually do that?

anyway, `docker` fails to parse such dockerfile:

```
$ docker build .
Sending build context to Docker daemon 4.096 kB
Error response from daemon: Syntax error - can't find = in "c". Must be of the form: name=value
$ cat Dockerfile
FROM scratch
LABEL a=b   c
```

dockerfile_parse/util.py
TomasTomecek: no, this is just to initialize the variable: the first condition in the loop will always take the else branch

twaugh: ..because a valid LABEL or ENV will always take a value, so shlex_splits_raw will always have at least one item, is that right?

TomasTomecek: That was my reasoning, but on the other hand, someone can easily parse a broken dockerfile: in such case, do we just raise a ValueError?

twaugh: I think ideally we'd raise ValueError with a descriptive message, yes.

TomasTomecek: I changed the initial value to "" as suggested by Luiz.

Unfortunately, `shlex.split` is too good. In order to make these crazy inputs work the same way as dockerd does, we would need change a lot of code. I'm not gonna do that in this PR. I don't think it's such a bad thing that we can parse more than dockerd. Worst case the build fails, no big deal.

tests/test_parser.py
```diff
@@ -694,3 +694,41 @@ def test_expand_concatenated_variables(self, dfparser):
...
+                  b="parsing is so hard" \
+                  c=y e s
+        """)
```
twaugh: This test case is not correct. Docker will not parse it. All labels in a statement have to be in the `k=v` form, or all of them have to be in `k v` form. You aren't allowed to mix and match within a single LABEL statement.
```
Sending build context to Docker daemon 2.048 kB
Error response from daemon: Syntax error - can't find = in "c". Must be of the form: name=value
```

tests/test_parser.py
```diff
@@ -694,3 +694,41 @@ def test_expand_concatenated_variables(self, dfparser):
...
+            "a": "b c d e",
+            "b": "parsing is so hard",
+            "c": "y e s"
+        }
+
+    def test_multiline2(self, dfparser):
+        dfparser.content = dedent("""
+            LABEL a="b c" \
+                  b=parsing is so hard \
+                  c="y e s"
```
twaugh: This test case is not valid because it mixes and matches the two forms of labels within a statement.
```
Sending build context to Docker daemon 2.048 kB
Error response from daemon: Syntax error - can't find = in "is". Must be of the form: name=value
```

tests/test_parser.py
```diff
@@ -694,3 +694,41 @@ def test_expand_concatenated_variables(self, dfparser):
...
+        """)
+        assert dfparser.labels == {
+            "a": "b c",
+            "b": "parsing is so hard",
+            "c": "y e s"
+        }
+
+    def test_multispace_label(self, dfparser):
+        # this fails with docker but succeeds with us b/c shlex.split eats consecutive spaces
+        dfparser.content = "LABEL a=b  c"
```
twaugh: Not valid due to mix-and-match.
```
Sending build context to Docker daemon 2.048 kB
Error response from daemon: Syntax error - can't find = in "c". Must be of the form: name=value
```

tests/test_parser.py
```diff
@@ -694,3 +694,41 @@ def test_expand_concatenated_variables(self, dfparser):
...
+        assert dfparser.labels == {"a": "b c"}
+
+    def test_broken_multiline_label(self, dfparser):
+        # once again, shlex.split is smarter than dockerd's parser
+        dfparser.content = dedent("""
+            LABEL a=b \
+                  c
```
twaugh: Not valid.
```
Sending build context to Docker daemon 2.048 kB
Error response from daemon: Syntax error - can't find = in "c". Must be of the form: name=value
```

twaugh:
Real test case should be:
```
FROM scratch
ENV VAR="foo bar"
LABEL LABEL=$VAR
```

Docker's parser is happy with this but dockerfile_parse is not.

twaugh:
I have a fix, will open a new PR for it.

twaugh:
See https://github.com/DBuildService/dockerfile-parse/pull/54

TomasTomecek:
feel free to close my PR then

PR #54, opened 2018-02-07
Fix "LABEL summary=$SUMMARY" handling

twaugh:
These commits add some new test cases for valid and invalid input, and implement the changes required to fix the failing tests by combining environment variable substitution with quote removal.

The EnvSubst class is renamed to WordSplitter, and provides two methods:
* dequote() returns the string with quotes/escaped consumed
* split() breaks the string into words at whitespace

This new class can be told whether or not to apply environment variable substitutions.

The way we process ENV/LABEL instructions is now to first break the instruction value into words, without quoting or applying environment variable substitutions. We can then inspect the first word to find out which key/value form is in use ("key=value" vs "key the value").

Based on this we now apply environment variable substitution to each key and value separately, consuming quotes and escape characters.

A similar process is used when modifying an ENV/LABEL instruction, with the exception that key=value forms are left alone with quotes and escapes intact unless modified.  This is the reason for changing some existing test cases.

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -17,7 +17,7 @@ from contextlib import contextmanager
 
 from .constants import DOCKERFILE_FILENAME
 from .util import (b2u, extract_labels_or_envs, get_key_val_dictionary,
-                   remove_quotes, shlex_split, u2b, Context)
+                   u2b, Context, WordSplitter)
 
 try:
     # py3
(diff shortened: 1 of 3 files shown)


tests/test_parser.py
```diff
@@ -438,16 +438,16 @@ def test_setter(self, dfparser, instruction, existing, new, expected):
...
         # Simple case, '=' but no quotes
         ('Release=1', 'Release', '6', 'Release=6'),
         # '=' and quotes
-        ('"Name"=\'alpha alpha\' Version=1', 'Name', 'beta delta', 'Name=\'beta delta\' Version=1'),
+        ('"Name"=\'alpha alpha\' Version=1', 'Name', 'beta delta', '"Name"=\'beta delta\' Version=1'),
         # '=', multiple labels, no quotes
         ('Name=foo Release=3', 'Release', '4', 'Name=foo Release=4'),
         # '=', multiple labels and quotes
-        ('Name=\'foo bar\' "Release"="4"', 'Release', '5', 'Name=\'foo bar\' Release=5'),
+        ('Name=\'foo bar\' "Release"="4"', 'Release', '5', 'Name=\'foo bar\' "Release"=5'),
```
rcerven: maybe add as well test:

('Release=1', 'Release', 'new value', "Release='new value'"),

rcerven: and 

('Release 1', 'Release', 'new value', "Release 'new value'")

twaugh: Thanks, this found a bug. :-)

twaugh: Just pushed another commit to fix this. Put it in its own commit for now for ease of review; will squash before merge.

dockerfile_parse/parser.py
```diff
@@ -443,26 +442,24 @@ def _modify_instruction_label_env(self, instruction, instr_key, instr_value):
                     endline = candidate['endline']
                     break
             else:  # LABEL/ENV "name"="value"
-                for index, token in enumerate(splits):
-                    words = token.split("=", 1)
-                    if words[0] == instr_key:
+                for index, token in enumerate(words):
+                    key_val = token.split("=", 1)
```
lcarva: Since we're asserting that length is 2, what about using: `key, val = token.split("=", 1)`

twaugh: Good idea. Added another commit to do that, which I'll squash before merging.

twaugh:
Squashed down to 2 commits ready for merge.

lcarva:
Updated release notes draft:
> Fix "LABEL summary=$SUMMARY" handling

PR #56, opened 2018-03-02
add newline if necessary when adding instructions

mlangsdorf:
If the existing lines do not end with \n, make sure they do.

Signed-off-by: Mark Langsdorf <mlangsdo@redhat.com>

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -528,6 +528,8 @@ class DockerfileParser(object):
             new_line = '{0} {1}\n'.format(instruction, value)
         if new_line:
             lines = self.lines
+            if not lines[len(lines) - 1].endswith('\n'):
+                new_line = '\n' + new_line
             lines += new_line
             self.lines = lines
 
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -748,3 +748,18 @@ class TestDockerfileParser(object):
         ])
         with pytest.raises(Exception):
             dfparser.labels
+
+    def test_remove_whitespace(self, tmpdir):
+        """
+        Verify keys are parsed correctly even if there is no final newline.
+
+        """
+        with open(os.path.join(str(tmpdir), 'Dockerfile'), 'w') as fp:
+            fp.write('FROM scratch')
+        tmpdir_path = str(tmpdir.realpath())
+        df1 = DockerfileParser(tmpdir_path)
+        df1.labels['foo'] = 'bar'
+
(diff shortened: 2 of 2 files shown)


tests/test_parser.py
```diff
@@ -748,3 +748,15 @@ def test_label_invalid(self, dfparser, value):
...
+        """
+        Verify keys are parsed correctly even if there is no final newline.
+
+        """
+        with open(os.path.join(str(tmpdir), 'Dockerfile'), 'wb') as fp:
+            fp.write('FROM scratch')
+        tmpdir_path = str(tmpdir.realpath())
+        df1 = DockerfileParser(tmpdir_path)
+        df1.labels['foo'] = 'bar'
+        assert 'scratchLABEL' not in df1.content
```
twaugh: Might be better to assert what we do expect, e.g. that if we make another DockerfileParser, a df2.from == scratch and df2.labels is as expected.

dockerfile_parse/parser.py
```diff
@@ -528,6 +528,9 @@ def _add_instruction(self, instruction, value):
             new_line = '{0} {1}\n'.format(instruction, value)
         if new_line:
             lines = self.lines
+            if not lines[len(lines) - 1].endswith('\n'):
+                newer_line = '\n' + new_line
+                new_line = newer_line
```
twaugh: Can just say `new_line = '\n' + new_line`.

tests/test_parser.py
mlangsdorf: Fixed.

dockerfile_parse/parser.py
mlangsdorf: Fixed.

tests/test_parser.py
```diff
@@ -748,3 +748,18 @@ def test_label_invalid(self, dfparser, value):
         ])
         with pytest.raises(Exception):
             dfparser.labels
+
+    def test_remove_whitespace(self, tmpdir):
+        """
+        Verify keys are parsed correctly even if there is no final newline.
+
+        """
+        with open(os.path.join(str(tmpdir), 'Dockerfile'), 'wb') as fp:
```
lcarva: Opening file in binary mode causes python3 to require bytes to be written to it. Just 'w' should do here since you are writing a text file.

mlangsdorf: Fixed.

lcarva:
Updated release notes draft:
> add newline if necessary when adding instructions

PR #58, opened 2018-05-31
Improve capabilities for multistage Dockerfiles

sosiouxme:
Introduce `add_lines` and `add_lines_at` methods for simpler manipulation.
Scope most existing properties to only the final stage.
Add `parent_images` property for getting and setting a list of images from the stages.
Introduce `is_multistage` property for convenience.

@twaugh @lcarva @mlangsdorf @rcerven PTAL
/cc @peterkline @adammhaile does this help...

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -14,6 +14,7 @@ import logging
 import os
 import re
 from contextlib import contextmanager
+from six import string_types
 
 from .constants import DOCKERFILE_FILENAME
 from .util import (b2u, extract_labels_or_envs, get_key_val_dictionary,
@@ -110,13 +111,12 @@ class DockerfileParser(object):
 
         self.env_replace = env_replace
 
-        if isinstance(parent_env, dict):
+        if parent_env is None:
+            self.parent_env = {}
+        else:
+            assert isinstance(parent_env, dict)
             logger.debug("Setting inherited parent image ENV vars: %s", parent_env)
             self.parent_env = parent_env
-        elif parent_env is not None:
-            assert isinstance(parent_env, dict)
-        else:
-            self.parent_env = {}
 
     @contextmanager
     def _open_dockerfile(self, mode):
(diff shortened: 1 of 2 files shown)


adammhaile:
I knew I was going to like having you on my team @sosiouxme 
Looks good :)

adammhaile:
@sosiouxme Python 2.6 only came out 10 years ago. Clearly it should be still tested :P

dockerfile_parse/parser.py
```diff
@@ -533,6 +581,85 @@ def _add_instruction(self, instruction, value):
...
+            df_lines[-1] = _endline(df_lines[-1])
+
+        # iterate through the stages in reverse order
+        # so adding lines doesn't invalidate line numbers from structure dicts.
+        # first add a bogus instruction to represent EOF in our iteration.
+        froms.append({'startline': len(df_lines) + 1})
+        for stage in range(len(froms)-2, -1, -1):  # e.g. 0 for single or 2, 1, 0 for 3 stages
+            start, finish = froms[stage], froms[stage+1]
+            linenum = start['endline'] + 1 if at_start else finish['startline']
+            df_lines[linenum:linenum] = [_endline(line) for line in lines]
```
twaugh: Nit: we're calculating _endline() repeatedly for the same values. Can we move that out of the loop?

tests/test_parser.py
```diff
@@ -203,6 +256,55 @@ def test_get_instructions_from_df(self, dfparser, instruction, instr_value,
...
+            CMD do some stuff
+
+            FROM {1}
+            COPY --from=builder some stuff
+            """)
+        dfparser.content = template.format(*FROM)
+
+        parents = dfparser.parent_images
+        for img in FROM:
+            assert img in parents
```
twaugh: How come we're not specifying the order they should appear?

dockerfile_parse/parser.py
```diff
@@ -533,6 +581,85 @@ def _add_instruction(self, instruction, value):
...
+        elif isinstance(anchor, dict):  # structure
+            assert anchor in self.structure, "Current structure does not match: {}".format(anchor)
+            if replace:
+                df_lines[anchor['startline']:anchor['endline'] + 1] = []
+            anchor = anchor['startline']
+        elif isinstance(anchor, string_types):  # line contents
+            matches = [index for index, text in enumerate(df_lines) if text == anchor]
+            if not matches:
+                raise RuntimeError("Cannot find line in the build file:\n" + anchor)
+            anchor = matches[-1]
```
twaugh: Oh, the *last* match? Can we have a test case that checks this?

dockerfile_parse/parser.py
```diff
@@ -266,44 +266,101 @@ def json(self):
...
+    def parent_images(self):
+        """
+        :return: list of parent images -- one image per each stage's FROM instruction
+        """
+        parents = []
+        for instr in self.structure:
+            if instr['instruction'] != 'FROM':
+                continue
+            image, _ = image_from(instr['value'])
+            if image is not None:
```
twaugh: I don't think we have testing that covers this not being the case.

tests/test_parser.py
sosiouxme: Thanks, I didn't update this with the implementation

dockerfile_parse/parser.py
sosiouxme: Modified test_add_lines_at to cover this

dockerfile_parse/parser.py
sosiouxme: Added one. And actually setting parent_images would be broken in that case. Well, everything would be broken anyway, but we can at least do no further harm.

lcarva:
Updated release notes draft:
> * Improve capabilities for multistage Dockerfiles

PR #60, opened 2018-06-04
test.sh: add resurrection, EXTRA_MOUNT, htmlcov

sosiouxme:
Have this reuse an existing container that's been stopped.
Also make it easy to mount more volumes into the test container.
Generate HTML coverage report.


lcarva:
Not updating release notes draft because this is a change solely to unit tests.

PR #64, opened 2018-07-13
Do not pin pytest/pytest-cov to specific versions

twaugh:


lcarva:
And somehow that breaks python 2.6 and 3.x tests...

twaugh:
Yes. I don't remember why we decided we needed to pin them -- probably to avoid this issue. But there must be another way...

twaugh:
I have no idea how to fix this.

PR #66, opened 2018-07-16
Require six package

twaugh:


lcarva:
I think this needs to be rebased to pick up centos fix.

twaugh:
Tests pass now.

PR #67, opened 2018-11-05
skip_scratch option for add_lines

rcerven:

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -597,11 +597,13 @@ class DockerfileParser(object):
         :param all_stages: bool for whether to add in all stages for a multistage build
                            or (by default) only the last.
         :param at_start: adds at the beginning (after FROM) of the stage(s) instead of the end.
+        :param skip_scratch: skip stages which use "FROM scratch"
         """
         assert len(lines) > 0
         lines = [_endline(line) for line in lines]
         all_stages = kwargs.pop('all_stages', False)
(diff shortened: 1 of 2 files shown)


twaugh:
See #68 for the fix for the test-case failure on Python 2.6.

dockerfile_parse/parser.py
```diff
@@ -597,11 +597,13 @@ def add_lines(self, *lines, **kwargs):
         :param all_stages: bool for whether to add in all stages for a multistage build
                            or (by default) only the last.
         :param at_start: adds at the beginning (after FROM) of the stage(s) instead of the end.
+        :param skip_scratch: skip for FROM scratch
```
twaugh: Would be nice to make this slightly more explicit, like: `skip stages which use "FROM scratch"`

rcerven: ok

rcerven:
we don't want to add lines with yum repos to FROM scratch sections

MartinBasti:
> we don't want to add lines with yum repos to FROM scratch sections

IIUC correctly, you are updating underlying library to not load `FROM scratch` sections, to prevent accidental updates of yum repos in wrong section in atomic-reactor(?), instead of handling this logic inside atomic-reactor(?) ?

rcerven:
rn updated:
* skip_scratch option for add_lines, to skip adding lines to FROM scratch sections

PR #78, opened 2019-05-02
initial implementation of comment instructions in DockerfileParser.structure

sebpiq:
Here is a first pass at implementing **COMMENT** instruction. 

- I decided to treat `value` as all the lines of the comment combined into a clean string, with "# " and new lines stripped. e.g. : 
```
# bla
#   blo
```
becomes 
```
'bla blo'
```
- I used private inlined functions in `structure` in order to avoid repetition, but this calls for a refactor. I didn't dare to do something too radical as I don't know how you see things should be. I'd recommend having `Instruction` and `Comment` as classes (e.g. subclasses of `dict`), and these inline functions could be methods.

Please let me know how this looks, I can push some fixes if needed.

### DIFF
--- a/dockerfile_parse/constants.py
+++ b/dockerfile_parse/constants.py
@@ -14,3 +14,4 @@ from sys import version_info
 PY2 = version_info[0] == 2
 
 DOCKERFILE_FILENAME = 'Dockerfile'
+COMMENT_INSTRUCTION = 'COMMENT'
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -20,6 +20,7 @@ from textwrap import dedent
 
 from dockerfile_parse import DockerfileParser
 from dockerfile_parse.parser import image_from
+from dockerfile_parse.constants import COMMENT_INSTRUCTION
 from tests.fixtures import dfparser, instruction
 
 NON_ASCII = "žluťoučký"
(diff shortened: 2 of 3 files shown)


dockerfile_parse/constants.py
```diff
@@ -14,3 +14,4 @@
 PY2 = version_info[0] == 2
 
 DOCKERFILE_FILENAME = 'Dockerfile'
+COMMENT_INSTRUCTION = 'COMMENT'
```
MartinBasti: Missing newline at the end of file

tests/test_parser.py
```diff
@@ -84,44 +85,65 @@ def test_constructor_cache(self, tmpdir):
...
+                          " label  foo  \\\n",          # extra ws
+                          "# multi-line \n",            # multi-line comment, interrupt instruction
+                          "# ...comment\n",
+                          "    bar  \n",                # extra ws, instruction continuation
                           "USER  {0}\n".format(NON_ASCII),
-                          "# comment \\\n",     # extra ws
-                          "# comment \\ \n",    # extra ws with a space
-                          "# comment \\\\ \n",  # two backslashes
+                          "# comment \\\n",             # extra ws
+                          "# with \\ \n",               # extra ws with a space
```
MartinBasti: I miss test cases for:
* `#comment` - comment without spaces
* `# comment with # inside`

tests/test_parser.py
```diff
@@ -84,44 +85,65 @@ def test_constructor_cache(self, tmpdir):
...
     def test_dockerfile_structure(self, dfparser):
-        dfparser.lines = ["# comment\n",        # should be ignored
-                          " From  \\\n",        # mixed-case
-                          "   base\n",          # extra ws, continuation line
-                          " # comment\n",
-                          " label  foo  \\\n",  # extra ws
-                          "    # comment\n",    # should be ignored
-                          "    bar  \n",        # extra ws, continuation line
+        dfparser.lines = ["# comment\n",                # single-line comment
+                          " From  \\\n",                # mixed-case
```
MartinBasti: IMHO we should also tests case
```
FROM \
   # comment
   fedora:30
```

tests/test_parser.py
```diff
@@ -84,44 +85,65 @@ def test_constructor_cache(self, tmpdir):
...
                                        'value': 'base'},
+                                      {'instruction': COMMENT_INSTRUCTION,
+                                       'startline': 3,
+                                       'endline': 3,
+                                       'content': ' #    another   comment\n',
+                                       'value': 'another   comment'},
+                                      {'instruction': COMMENT_INSTRUCTION,
+                                       'startline': 5,
+                                       'endline': 6,
+                                       'content': '# multi-line \n# ...comment\n',
```
MartinBasti: Docker has NO multi-line comments. Why we should handle comments like multi-line in parser? I vote for having all comments as single line, there is no format support for multiline comments in Dockerfile

sebpiq: Reasoning behind this is that you usually consider several contiguous lines of comments as one comment block. If we remove the multi-line comments, then imagine a very long documentation block - say 20 lines of comments - will be parsed as 20 separate "instruction" dictionaries, which feels a bit weird. Though I can see that supporting multi-lines comment is weird too ... It's your call really I don't care. Let me know and I'll change this.

tests/test_parser.py
sebpiq: :+1:

tests/test_parser.py
sebpiq: It's already there (see just below "interrupt instruction"), though ATM it is 2 test cases in one (multi-line comment + comment interrupting a multi-line instruction)

sebpiq: It's already there (see just below "interrupt instruction"), though ATM it is 2 test cases in one (multi-line comment + comment interrupting a multi-line instruction)

dockerfile_parse/constants.py
sebpiq: :+1:

tests/test_parser.py
MartinBasti: Ahh, sorry for the noise I haven't noticed that

tests/test_parser.py
MartinBasti: I'd stay with oneline comments only but let's wait for other opinions here

sebpiq: Ok since there are no other opinions, I guess I'll just remove the multi-line comments

sebpiq-kialo:
@MartinBasti sorry for the delay. Implemented the changes requested. Could you have another look please?

MartinBasti:
Well, parser directives are comments. According 
```
The unknown directive is treated as a comment due to not being recognized. In addition, the known directive is treated as a comment due to appearing after a comment which is not a parser directive.

# unknowndirective=value
# knowndirective=value
```
we would need to know all supported parser directives to differentiate between parser directive and comment.
I'd stay with comment only now.

MartinBasti:
@sebpiq could you please address that nitpicks?

sebpiq:
@MartinBasti I rewrote the commit message, but not sure what @athos-ribeiro wants, as there are already both tests for interrupted and non-interrupted multi-line instructions. Sure one of these is a `RUN` and the other is a `LABEL` but the parser doesn't make the distinction so IMO there is no point in adding another test case.

MartinBasti:
> but the parser doesn't make the distinction

For now but it may change in future a we want to avoid that. `RUN` is very important.

dockerfile_parse/constants.py
twaugh: Any reason not to make this `'#'`?

sebpiq: yeah I about it but thought `'COMMENT'` was more self-explanatory. Either way it's arbitrary.

MartinBasti:
Release notes:
> * implementation of comment instructions in DockerfileParser.structure

PR #80, opened 2019-07-08
enable fedora 30 testing

rcerven:
and disable fedora 28 testing, build py2 packages only for
rhel <=7, and py3 for fedora and rhel >= 8


rcerven:
rn updated: enable building py3 package for rhel8

PR #82, opened 2019-07-12
Give a better error message for ENV/LABEL syntax error

chmeliik:
* OSBS-7401

Previously, instructions such as "LABEL a=1 b" would cause a tuple
unpacking exception, now the error message is the same as what you would
get from docker.

### DIFF
--- a/dockerfile_parse/util.py
+++ b/dockerfile_parse/util.py
@@ -239,6 +239,10 @@ def extract_labels_or_envs(env_replace, envs, instruction_value):
         #   LABEL/ENV "name"="value" ["name"="value"...]
         # Each word is a key=value pair.
         for k_v in words:
+            if '=' not in k_v:
+                raise ValueError('Syntax error - can\'t find = in "{word}". '
+                                 'Must be of the form: name=value'
+                                 .format(word=k_v))
             key, val = [substitute_vars(x) for x in k_v.split('=', 1)]
             key_val_list.append((key, val))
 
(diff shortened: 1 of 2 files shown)


dockerfile_parse/util.py
```diff
@@ -239,6 +239,10 @@ def substitute_vars(val):
         #   LABEL/ENV "name"="value" ["name"="value"...]
         # Each word is a key=value pair.
         for k_v in words:
+            if '=' not in k_v:
```
MartinBasti: I'm curios about cases:
* LABEL key=inside_key=value
* LABEL "key=value"
* LABEL something"key=value"

MartinBasti: and LABEL key==value

chmeliik: All of those are valid LABELs according to docker:
```Dockerfile
FROM fedora:30

LABEL key=inside_key=value
LABEL "key2=value"
LABEL key"3=value"
LABEL key4==value
```

```bash
$ docker build -t testimage .
$ docker inspect testimage
```

```json
...

"Labels": {
    "key": "inside_key=value",
    "key2": "value",
    "key3": "value",
    "key4": "=value"
}
...
```

MartinBasti: Ok, thank you!

chmeliik:
Rebased on master.

MartinBasti:
Release notes

> * improved error reporting when incorrect LABEL syntax is used

PR #88, opened 2019-09-24
Extend unicode testing

MartinBasti:
I extended string in current tests with '❤' character to test Unicode
non-ASCII characters (utf-8 encoded)

* OSBS-4457

Signed-off-by: Martin Bašti <mbasti@redhat.com>

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -186,22 +186,22 @@ class TestDockerfileParser(object):
     def test_dockerfile_json(self, dfparser):
         dfparser.content = dedent("""\
             # comment
-            From  base
-            LABEL foo="bar baz"
+            From  base:❤
+            LABEL foo="bar❤baz"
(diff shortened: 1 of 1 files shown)


MartinBasti:
Instead of duplicating tests for unicode testing, I chose to update current tests with unicode character `❤`
Let me know what do you think

MartinBasti:
I'm not sure about testing FROM instruction with unicode, because:
```
$ podman pull 'fedora:❤'                                                                         
Error: error pulling image "fedora:❤": unable to pull fedora:❤: error getting default registries to try: invalid reference format
```

twaugh:
I don't see any spec that forbids UTF-8 in image tag names.

MartinBasti:
I updated 2 simple cases to test `FROM` instruction

MartinBasti:
no RN: only unitests updated

PR #94, opened 2020-01-24
Add ARG parsing

### LINKED ISSUE #74: ARGs are not handled in FROM instruction
I have the following Dockerfile (https://docs.docker.com/v17.09/engine/reference/builder/#understand-how-arg-and-from-interact):

    ARG PYTHON_VERSION
    FROM python:${PYTHON_VERSION}

When trying to parse it with this library:

    >>> parser = dockerfile_parse.DockerfileParser('Dockerfile')
    >>> parser.parent_images
    ['python:${PYTHON_VERSION}']

What I expected:
- `build_args` constructor argument (similar to parent_env)
- args to be expanded in `parent_images`

nishakm:
Added a new property "args" which gets key-value pairs for ARG
instructions in Dockerfiles. This change adds a getter for args.
Also added environment variable replacement in the parent_images
property.

Resolves #74

Signed-off-by: Nisha K <nishak@vmware.com>

- [ ] "Signed-off-by:" line is present in each commit

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -70,6 +70,26 @@ class Envs(dict):
         self.parser.envs = dict(self)
 
 
+class Args(dict):
+    """
+    A class for allowing direct write access to Dockerfile args, e.g.:
+
+    parser.args['variable_name'] = 'value'
+    """
+
+    def __init__(self, args, parser):
+        super(Args, self).__init__(args)
+        self.parser = parser
+
+    def __delitem__(self, key):
+        super(Args, self).__delitem__(key)
+        self.parser.envs = dict(self)
+
+    def __setitem__(self, key, value):
+        super(Args, self).__setitem__(key, value)
+        self.parser.envs = dict(self)
+
+
 class DockerfileParser(object):
     def __init__(self, path=None,
                  cache_content=False,
@@ -243,7 +263,7 @@ class DockerfileParser(object):
         insnre = re.compile(r'^\s*(\S+)\s+(.*)$')  # matched group is insn
         contre = re.compile(r'^.*\\\s*$')          # line continues?
         commentre = re.compile(r'^\s*#')           # line is a comment?
-        
+
         in_continuation = False
         current_instruction = None
 
@@ -265,7 +285,7 @@ class DockerfileParser(object):
                     if not m:
                         continue
                     current_instruction = _create_instruction_dict(
-                        instruction=m.groups()[0].upper(), 
+                        instruction=m.groups()[0].upper(),
                         value=_rstrip_backslash(m.groups()[1])
                     )
                 else:
@@ -301,6 +321,10 @@ class DockerfileParser(object):
                 continue
             image, _ = image_from(instr['value'])
             if image is not None:
+                # replace any ARG keys with values
+                for key in self.args.keys():
+                    pattern = '\\$'+key+'|'+'\\${'+key+'}'
+                    image = re.sub(pattern, self.args[key], image)
                 parents.append(image)
         return parents
 
@@ -411,6 +435,14 @@ class DockerfileParser(object):
         """
         return self._instruction_getter('ENV', env_replace=self.env_replace)
 
+    @property
+    def args(self):
+        """
+        ARGs from Dockerfile
+        :return: dictionary of arg_var_name:value (value might be '')
+        """
+        return self._instruction_getter('ARG', env_replace=self.env_replace)
+
     def _instruction_getter(self, name, env_replace):
         """
         Get LABEL or ENV instructions with environment replacement
@@ -419,7 +451,8 @@ class DockerfileParser(object):
         :param env_replace: bool, whether to perform ENV substitution
         :return: Labels instance or Envs instance
         """
-        if name != 'LABEL' and name != 'ENV':
+        supported_instrs = ('LABEL', 'ENV', 'ARG')
+        if name not in supported_instrs:
             raise ValueError("Unsupported instruction '%s'", name)
         instructions = {}
         envs = {}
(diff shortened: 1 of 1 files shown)


nishakm:
Hi! Sorry it took so long. I'm not sure what tests I need to run locally so this change is acceptable. Any guidance regarding that would be appreciated :)

chmeliik:
Hi, thanks for the contribution!

To run tests, please use the `test.sh` script - just run `./test.sh` from the dockerfile-parse directory. This will run tests for python 2 in a centos:7 docker image. You can also run tests for other OS/python combinations like so: `OS=fedora OS_VERSION=30 PYTHON_VERSION=3 ./test.sh` (see the Travis build details for what combinations are tested).

Once you get those working, please also add tests for your changes :blush:

dockerfile_parse/parser.py
```diff
@@ -301,6 +321,10 @@ def parent_images(self):
                 continue
             image, _ = image_from(instr['value'])
             if image is not None:
+                # replace any ARG keys with values
```
MartinBasti: I don't think that this is proper solution. ARGs may be used on many places, you covered only one case.
Also parser should return exact content of dockerfile if not specified otherwise.

I'd go with a separate method which replaces ARGs within dockerfile and returns new parser object with replaced occurrences of ARGS.

WDYT?

dockerfile_parse/parser.py
```diff
@@ -442,7 +474,12 @@ def _instruction_getter(self, name, env_replace):
                         envs[key] = value
 
         logger.debug("instructions: %r", instructions)
-        return Labels(instructions, self) if name == 'LABEL' else Envs(instructions, self)
+        if name == 'LABEL':
+            return Labels(instructions, self)
+        elif name == 'ENV':
+            return Envs(instructions, self)
+        else:
```
MartinBasti: Would be possible to make it explicit `elif name == 'ARG':`  and keep compatibility with `else` branch returning `Envs`?

dockerfile_parse/parser.py
MartinBasti: Respectively for ENVs there is:
```
```
in init. We can follow the same logic as was used for envs with args as they are basically the same only used in different context.

chmeliik: > I don't think that this is proper solution. ARGs may be used on many places, you covered only one case.

To be fair, it's not like the current implementation supports env replacement anywhere other than the `envs` and `labels` properties. If we want to support all instructions, it might be best to use a different approach (such as modifying the `_instruction_getter()` method to support every type of instruction or incorporating the `context_structure` property).

I'm not sure if the intention of this PR was to support env/arg replacement in FROM instructions but support for ARG handling will nevertheless need some more work.

nishakm: Agree that this is a quick n' dirty replacement in FROM mostly to address the specific issue in #74 that ARG values are not replaced in FROM. If you would rather merge in a PR that supports ARG completey I could work on that.

MartinBasti: I'd like to have ARGs and ENVs aligned handling. No need for extra functionality which is not supported by ENVs already (Except that FROM replacement).

nishakm: > I don't think that this is proper solution. ARGs may be used on many places, you covered only one case.
> Also parser should return exact content of dockerfile if not specified otherwise.
> 
> I'd go with a separate method which replaces ARGs within dockerfile and returns new parser object with replaced occurrences of ARGS.
> 
> WDYT?

What I am noticing is that env replacement happens just for ENV or LABEL instructions and not for RUN instructions nor anywhere in the dockerfile content itself. So I implemented the ARG getter the same way as the ENV getter which uses `_instruction_getter()`. The replacement in the FROM line is to address the specific issue reported in #74. I can do a separate PR for global content replacement which will remove this change. Or maybe I'm missing something in the implementation...

nishakm: > Respectively for ENVs there is:
> 
> ```
> self.env_replace = env_replace
> ```
> 
> in init. We can follow the same logic as was used for envs with args as they are basically the same only used in different context.

Taking a step back here: What does `env_replace` do? It doesn't say in the docstring.

chmeliik: `env_replace` is a flag which decides if ENV variables will be  expanded or not. However, as you rightly pointed out, this variable expansion currently only supports ENV and LABEL instructions.

I think what we want is for ARGs to be treated the same as ENVs, as in support ARG expansion wherever ENV expansion is currently supported etc. I'm afraid you will find a lot of the logic required for that is not yet in place though.

If you would rather only support ARG expansion in FROM instructions then that is fine too, as long as it respects the `env_replace` flag (or a new one specifically for this) and is properly documented.

nishakm: @MartinBasti @chmeliik Sorry for the late reply.
The trouble with maintaining the same logic for ARGs and ENVs with variable expansion is that ARGs are not used the same way as ENVs. If I apply the same ENVs expansion to ARGs in the Dockerfile examples from https://docs.docker.com/engine/reference/builder/#arg
they end up looking like this:
```
# from example: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
ARG VERSION=latest
FROM busybox:$VERSION
ARG VERSION
RUN echo $VERSION > image_version
 
if env_replace is True:
-> {'VERSION': 'latest'}

if env_replace is False:
-> {'VERSION': 'latest', 'VERSION': ''} (Is this expected behaviour?)
```
It also seems to me you don't want to fix #74 (ARG replacement in the FROM instruction)?

MartinBasti: Sorry for late reply. I'm not exactly sure what you mean. What is difference between ARGs and ENVs except that ARGs are available only during buildtime

nishakm: Unlike ENVs, ARGs don't have to have a value for the variable in the Dockerfile (see `ARG VERSION` in the example). I am trying to understand what the expected behaviour is in the case where it doesn't have a variable and `env_replace` is True or False.

twaugh: ENVs also don't necessarily have a value defined in the Dockerfile -- it might come from the parent image's config. They feel quite similar to me.

durd07:
Hi, when will this be merged?

nishakm:
Closing this as another PR has been opened with more changes.

PR #95, opened 2020-02-17
[.packit.yaml] remove no longer needed keys:values

jpopelka:
[Some keys](https://packit.dev/docs/configuration/#top-level-keys) in the Packit config file are no longer mandatory since the default values are just fine.
One can also use [aliases](https://packit.dev/docs/configuration/#supported-jobs) for dist-git branches and/or copr build targets.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,14 +1,5 @@
 # https://packit.dev/docs/configuration/
 
-specfile_path: python-dockerfile-parse.spec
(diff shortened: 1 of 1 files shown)


MartinBasti:
it seems that we may need to replace rhel8 beta with epel8

MartinBasti:
Thank you @jpopelka !

PR #98, opened 2020-04-01
Add support for custom tito builder

MartinBasti:
Tito cannot handle if release tags are different from {name}-{version},
thus custom builder is needed.

Signed-off-by: Martin Bašti <mbasti@redhat.com>


MartinBasti:
no release notes, release change only

PR #100, opened 2020-04-14
Add pylint testing

SalatskySal:

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,76 @@
+[MASTER]
+
+# Use multiple processes to speed up Pylint.
+jobs=1
+
+# Pickle collected data for later comparisons.
+persistent=no
+
+# Activate the evaluation score.
+score=no
+
+[MESSAGES CONTROL]
+
+# Only show warnings with the listed confidence levels. Leave empty to show
+# all. Valid levels: HIGH, INFERENCE, INFERENCE_FAILURE, UNDEFINED
+confidence=
+
+# Disable the message, report, category or checker with the given id(s). You
+# can either give multiple identifiers separated by comma (,) or put this
+# option multiple times (only on the command line, not in the configuration
+# file where it should appear only once).You can also use "--disable=all" to
+# disable everything first and then reenable specific checks. For example, if
+# you want to run only the similarities checker, you can use "--disable=all
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use"--disable=all --enable=classes
+# --disable=W"
+enable=all,
+       python3
+disable=I,
+        R,  # refactoring checks
+        arguments-differ,  # nice to have
+        bad-continuation,  # pep8
+        bad-option-value,
+        bad-whitespace,  # pep8, nice to have
+        comprehension-escape,
+        import-outside-toplevel,
+        invalid-name,
+        len-as-condition,
+        missing-class-docstring,
+        missing-docstring,
+        missing-function-docstring,
+        no-self-use,
+        protected-access,
+        redefined-outer-name,
+        too-many-lines,
+        wrong-import-order,
+        wrong-import-position
+
+
+[REPORTS]
+
+# Template used to display messages. This is a python new-style format string
+# used to format the message information. See doc for all details
+msg-template='{path}:{line}: [{msg_id}({symbol}), {obj}] {msg})'
+
+# Set the output format. Available formats are text, parseable, colorized, json
+# and msvs (visual studio).You can also give a reporter class, eg
+# mypackage.mymodule.MyReporterClass.
+output-format=colorized
+
+[TYPECHECK]
+
+# List of class names for which member attributes should not be checked (useful
+# for classes with dynamically set attributes). This supports the use of
+# qualified names.
+ignored-classes=requests.structures.LookupDict
+
+# List of module names for which member attributes should not be checked
+# (useful for modules/projects where namespaces are manipulated during runtime
+# and thus existing member attributes cannot be deduced by static analysis. It
+# supports qualified module names, as well as Unix pattern matching.
+ignored-modules=requests.packages,
+                responses,
+                docker,
+                http.client,
+                six.moves
--- a/dockerfile_parse/__init__.py
+++ b/dockerfile_parse/__init__.py
@@ -10,6 +10,7 @@ of the BSD license. See the LICENSE file for details.
 # This file is generated from template: rel-eng/version_init.template
 #
 
+from __future__ import absolute_import
 from .parser import DockerfileParser
 
 __version__ = "0.0.16"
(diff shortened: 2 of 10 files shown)


.pylintrc
```diff
@@ -0,0 +1,76 @@
...
+
+# List of module names for which member attributes should not be checked
+# (useful for modules/projects where namespaces are manipulated during runtime
+# and thus existing member attributes cannot be deduced by static analysis. It
+# supports qualified module names, as well as Unix pattern matching.
+ignored-modules=requests.packages,
+                responses,
+                docker,
+                http.client,
+                six.moves
```
MartinBasti, line 76: missing newline

dockerfile_parse/__init__.py
```diff
@@ -10,6 +10,7 @@
 # This file is generated from template: rel-eng/version_init.template
 #
 
+from __future__ import absolute_import
 from .parser import DockerfileParser
```
chmeliik, line 14: The comment says this file is generated from template, maybe the import should be added to the template instead? Or just silence the warning, since the import is relative anyway

dockerfile_parse/parser.py
```diff
@@ -628,7 +628,7 @@ def add_lines(self, *lines, **kwargs):
         all_stages = kwargs.pop('all_stages', False)
         at_start = kwargs.pop('at_start', False)
         skip_scratch = kwargs.pop('skip_scratch', False)
-        assert not kwargs, "Unknown keyword argument(s): {0}".format(kwargs.keys())
+        assert not kwargs, "Unknown keyword argument(s): {0}".format(list(kwargs.keys()))
```
chmeliik, line 633: This could just be `list(kwargs)`, avoids double copy on python 2

dockerfile_parse/parser.py
```diff
@@ -667,12 +667,12 @@ def add_lines_at(self, anchor, *lines, **kwargs):
...
         after = kwargs.pop('after', False)
         assert not (after and replace)
-        assert not kwargs, "Unknown keyword argument(s): {0}".format(kwargs.keys())
+        assert not kwargs, "Unknown keyword argument(s): {0}".format(list(kwargs.keys()))
 
         # find the line number for the insertion
         df_lines = self.lines
         if isinstance(anchor, int):  # line number, just validate
-            assert anchor in range(len(df_lines))
+            assert anchor in list(range(len(df_lines)))
```
chmeliik, line 677: Is there a better way to fix this warning? I'm pretty sure `X in range` is O(1) on python 3 but `X in list` is O(n). Maybe just silence it, or use a good old `0 <= x < ...`

dockerfile_parse/parser.py
```diff
@@ -271,6 +271,8 @@ def _clean_comment_line(line):
                 else:
                     current_instruction['content'] += line
                     current_instruction['endline'] = lineno
+
+                    # pylint: disable=unsupported-assignment-operation
                     if current_instruction['value']:
                         current_instruction['value'] += _rstrip_backslash(line)
                     else:
```
chmeliik, line 278: Careful with pylint disable scope: http://pylint.pycqa.org/en/latest/user_guide/message-control.html
You generally want to disable the warning in the smallest scope possible, here it should be inside the if block

tests/test_parser.py
```diff
@@ -297,6 +297,7 @@ def test_get_parent_env_from_scratch(self, tmpdir):
 
         assert not df1.envs
 
+    # pylint: disable=anomalous-backslash-in-string
     @pytest.mark.parametrize(('instr_value', 'expected'), [
         ('"name1"=\'value 1\' "name2"=myself name3=""',
          {'name1': 'value 1',
```
chmeliik, line 303: I think this would disable the warning for the whole file from this line onwards.
It *might* work if you put it here instead:
```python
@pytest.mark.parametrize(('instr_value', 'expected'), [
    # pylint: disable=anomalous-backslash-in-string
    ('"name1"=\'value 1\' "name2"=myself name3=""',
```

dockerfile_parse/__init__.py
MartinBasti: Yes, add it into template too

dockerfile_parse/parser.py
MartinBasti: `X in range is O(1)` I'm not sure about this, it's still hidden iteration so I'd say O(n). But faster than alocating space for list, creating list and iterating list for sure.

dockerfile_parse/parser.py
SalatskySal: I tried to do it, but it _doesn't_ work. If I move disabling comment into if-statement, it will disable the error only for assignments in if-block, not else-block

SalatskySal: But thanks for the link. I'll be careful with this

dockerfile_parse/__init__.py
SalatskySal: Thanks, I didn't get attention for that :(

dockerfile_parse/parser.py
MartinBasti: you can re-enable that check after, if there is no other way:
```python
# pylint: disable=unsupported-assignment-operation
nananana
# pylint: enable=unsupported-assignment-operation
```

dockerfile_parse/parser.py
chmeliik: You could do something like
```python
def in_range(n, r):
    return r.start <= n < r.stop and ((n - r.start) % r.step) == 0
```
I'm sure if I thought of this, python people did too :D

dockerfile_parse/parser.py
chmeliik: Oh, the else statement is also an `unsupported-assignment-operation`? I thought it was just the `+=`. Oh well, this works :)

dockerfile_parse/parser.py
MartinBasti: if you are sure, you can link cpython code here :D

MartinBasti: @chmeliik kudos
https://github.com/python/cpython/blob/master/Objects/rangeobject.c#L405

chmeliik: @MartinBasti oh you actually found it, nice :D

MartinBasti: reducted kudos
it iterates sometimes: https://github.com/python/cpython/blob/master/Objects/rangeobject.c#L419

dockerfile_parse/util.py
```diff
@@ -24,8 +24,11 @@ def b2u(string):
 
 def u2b(string):
     """ unicode to bytes"""
-    if ((PY2 and isinstance(string, unicode)) or
-        ((not PY2) and isinstance(string, str))):
+    # pylint: disable=undefined-variable
+    # pylint: disable=unicode-builtin
+    string_type = unicode if PY2 else str
+
+    if isinstance(string, string_type):
```
tkdchen, line 31: Why not just simply use `six.text_type`?

SalatskySal: Wow, I didn't know about this feature, thanks :+1:

tkdchen:
LGTM

MartinBasti:
no release notes, internal change only

PR #102, opened 2020-04-19
Fix 'invalid escape sequence' warning

amezin:
When running tests on py 3.8, I get the following warning:

  /home/amezin/dockerfile-parse/dockerfile_parse/parser.py:236: DeprecationWarning: invalid escape sequence \s
    line = re.sub('^\s*#\s*', '', line)

Apparently, this should be a raw literal


MartinBasti:
We have to wait for packit https://github.com/packit-service/packit-service/issues/572

PR #107, opened 2020-06-14
Complete support for ARG statement handling

iciclespider:
This started with the branch @nishakm submitted, but adds complete support for ARGs usage in all the same places ENVs are currently supported. It also enhanced all ENV pytests to also test ARGs, and ensured pytests have 100% coverage.

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -15,79 +15,71 @@ import os
 import re
 from contextlib import contextmanager
 from six import string_types
+from six.moves import shlex_quote as quote
 
 from .constants import DOCKERFILE_FILENAME, COMMENT_INSTRUCTION
-from .util import (b2u, extract_labels_or_envs, get_key_val_dictionary,
+from .util import (b2u, extract_key_values, get_key_val_dictionary,
                    u2b, Context, WordSplitter)
-
-try:
-    # py3
-    from shlex import quote
-except ImportError:
-    from pipes import quote
 
 
 logger = logging.getLogger(__name__)
 
 
-class Labels(dict):
+class KeyValues(dict):
+    """
+    Abstract base class for allowing direct write access to Dockerfile
+    instructions which result in a set of key value pairs.
+
+    Subclasses must override the `parser_attr` value.
+    """
+    parser_attr = None
+
+    def __init__(self, key_values, parser):
+        super(KeyValues, self).__init__(key_values)
+        self.parser = parser
+
+    def __delitem__(self, key):
+        super(KeyValues, self).__delitem__(key)
+        setattr(self.parser, self.parser_attr, dict(self))
+
+    def __setitem__(self, key, value):
+        super(KeyValues, self).__setitem__(key, value)
+        setattr(self.parser, self.parser_attr, dict(self))
+
+    def __eq__(self, other):
+        if not isinstance(other, dict):
+            return False
+        return dict(self) == other
+
+    def __hash__(self):
+        return hash(json.dumps(self, separators=(',', ':'), sort_keys=True))
+
+
+class Labels(KeyValues):
     """
     A class for allowing direct write access to Dockerfile labels, e.g.:
 
     parser.labels['label'] = 'value'
     """
-
-    def __init__(self, labels, parser):
-        super(Labels, self).__init__(labels)
-        self.parser = parser
-
-    def __delitem__(self, key):
-        super(Labels, self).__delitem__(key)
-        self.parser.labels = dict(self)
-
-    def __setitem__(self, key, value):
-        super(Labels, self).__setitem__(key, value)
-        self.parser.labels = dict(self)
+    parser_attr = 'labels'
 
 
-class Envs(dict):
+class Envs(KeyValues):
     """
     A class for allowing direct write access to Dockerfile env. vars., e.g.:
 
     parser.envs['variable_name'] = 'value'
     """
-
-    def __init__(self, envs, parser):
-        super(Envs, self).__init__(envs)
-        self.parser = parser
-
-    def __delitem__(self, key):
-        super(Envs, self).__delitem__(key)
-        self.parser.envs = dict(self)
-
-    def __setitem__(self, key, value):
-        super(Envs, self).__setitem__(key, value)
-        self.parser.envs = dict(self)
+    parser_attr = 'envs'
 
 
-class Args(dict):
+class Args(KeyValues):
     """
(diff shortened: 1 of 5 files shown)


dockerfile_parse/parser.py
```diff
@@ -360,7 +395,14 @@ def baseimage(self, new_image):
...
         """
-        images = self.parent_images or [None]
+        images = []
+        for instr in self.structure:
+            if instr['instruction'] == 'FROM':
+                image, _ = image_from(instr['value'])
+                if image is not None:
+                    images.append(image)
+        if not images:
+            raise RuntimeError('No stage defined to set base image on')
```
MartinBasti, line 405: Why this  change is needed?

dockerfile_parse/parser.py
```diff
@@ -15,72 +15,87 @@
...
+    """
+    parser_attr = 'args'
 
 
 class DockerfileParser(object):
     def __init__(self, path=None,
                  cache_content=False,
                  env_replace=True,
                  parent_env=None,
+                 build_args=None,
```
MartinBasti, line 90: This has to be appended as last param, otherwise we may break BW compatibility for user who are just using it as positional args (I'd wish having `*` in params list in py2)

dockerfile_parse/util.py
```diff
@@ -46,13 +46,16 @@ class WordSplitter(object):
     SQUOTE = "'"
     DQUOTE = '"'
 
-    def __init__(self, s, envs=None):
+    def __init__(self, s, args=None, envs=None):
```
MartinBasti, line 49: Waiving BW compat, this is just util module

test.sh
```diff
@@ -71,7 +69,7 @@ function setup_dfp() {
 
   # CentOS needs to have setuptools updates to make pytest-cov work
   if [[ $OS != "fedora" ]]; then
-    $RUN $PIP install -U setuptools
+    $RUN $PIP install -U 'setuptools<45'
```
ben-alkov, line 72: You don't have to pin to a specific version here. As long as you

```bash
  $RUN $PIP install -U pip
  $RUN $PIP install -U setuptools
```

pip and setuptools will do the right thing

iciclespider:
I added a `test-test.sh` file which exercises all uses of `test.sh` to verify that my `test.sh` changes complete.

test.sh
MartinBasti: because old pip cannot understand supported python version listed in packages. Upgraded pip will install setuptools=44 under py2 as expected

test.sh
```diff
@@ -94,7 +99,7 @@ case ${ACTION} in
   ;;
 "pylint")
   setup_dfp
-  $RUN $PKG install -y "${PYTHON}-pylint"
+  $RUN $PIP install pylint
```
MartinBasti, line 102: We used distro specific version intentionally. Upstream pylint changes too often and causes many false positives we have to handle.

test-test.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
MartinBasti, line 1: Please drop this script from PR.

If needed travis CI should be updated, there are specific requirements, we don't test everything.

* latest bandit once per py2/py3 (it's not needed for every distro)
* markdownlint is enough to run just once too
* pylint py2, freezed at F29 as newer pylint version stopped supporting py2
* pylint py3, tested on latest fedora only, no need to test each distro
* F32 should be able run only py3, we will replace F30 with F32 soon

test.sh
ben-alkov: @iciclespider I just had to handle this exact issue in [osbs-client](https://github.com/containerbuildsystem/osbs-client/pull/984/files) - as long as you `$RUN $PIP install -U pip`, pip will do the right thing and install[1] a correct, working setuptools for the OS/Python it's running under.

[1] I did `$RUN $PIP install -U setuptools` immediately after updating pip; YMMV

test.sh
iciclespider: > Upstream pylint changes too often and causes many false positives we have to handle.

Would not this be better addressed by pip installing a specific version, rather then relying on the fact that the distro packaging is somewhat behind the latest?

test-test.sh
iciclespider: Removed.

test.sh
iciclespider: Oh, I see what you are saying. Added the pip upgrade before the setuptools upgrade.

test.sh
MartinBasti: It happens that when installing from pip some dependencies were broken, or incompatible dependencies were installed. We would need to freeze pip versions. So for me it's easier to rely on distro installations that resolves dependency issues for me.

MartinBasti:
Can you please squash last 2 commits into `Complete support for ARG statement handling` commit please?

python-dockerfile-parse.spec
```diff
@@ -121,7 +121,7 @@ py.test-%{python3_version} -v tests
 * Tue Jun 02 2020 Robert Cerven <rcerven@redhat.com> 0.0.18-1
 - new upstream release: 0.0.18
 
-* Fri Apr 24 2020 Martin Bašti <mbasti@redhat.com> 0.0.17-1
+* Fri Apr 24 2020 Martin Basti <mbasti@redhat.com> 0.0.17-1
```
MartinBasti, line 124: Pls add this change to separate commit

test.sh
```diff
@@ -23,6 +24,16 @@ for dir in ${EXTRA_MOUNT:-}; do
   engine_mounts=("${engine_mounts[@]}" -v "$dir":"$dir":z)
 done
 
+# Force recreation of the container
```
MartinBasti, line 27: test.sh changes are also unrelated to feature, can you extract it into separate commit?

iciclespider: By "separate commit", do you mean a different Pull Request? Or a separate commit within this Pull Request?

python-dockerfile-parse.spec
iciclespider: By "separate commit", do you mean a different Pull Request? Or a separate commit within this Pull Request?

iciclespider:
I reverted the `test.sh` changes to be the bare minimum needed to get this PR to pass the Travis CI build.

MartinBasti:
@athos-ribeiro @chmeliik @rcerven @ben-alkov PTAL

MartinBasti:
Release notes:
- Added support for ARG statement handling

PR #110, opened 2020-09-18
Handle escape directive

### LINKED ISSUE #109: Handle escape directive for Dockerfiles
**Problem description**
For Windows Dockerfiles it can often be beneficial to use the `escape` directive ([Documentation](https://docs.docker.com/engine/reference/builder/#escape)) as Windows uses `\` for paths. The current parser implementation does not handle this directive and reports errors on specific operations:

```
In [11]: df.context_structure
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-11-90b26318ceeb> in <module>
----> 1 df.context_structure

c:\python38\lib\site-packages\dockerfile_parse\parser.py in context_structure(self)
    818
    819             if instruction_type in ('ARG', 'ENV', 'LABEL'):
--> 820                 values = get_key_val_dictionary(
    821                     instruction_value=instr['value'],
    822                     env_replace=instruction_type != 'ARG' and self.env_replace,

c:\python38\lib\site-packages\dockerfile_parse\util.py in get_key_val_dictionary(instruction_value, env_replace, args, envs)
    256     args = args or {}
    257     envs = envs or {}
--> 258     return dict(extract_key_values(instruction_value=instruction_value,
    259                                    env_replace=env_replace,
    260                                    args=args, envs=envs))

c:\python38\lib\site-packages\dockerfile_parse\util.py in extract_key_values(env_replace, args, envs, instruction_value)
    244         for k_v in words:
    245             if '=' not in k_v:
--> 246                 raise ValueError('Syntax error - can\'t find = in "{word}". '
    247                                  'Must be of the form: name=value'
    248                                  .format(word=k_v))

ValueError: Syntax error - can't find = in "`". Must be of the form: name=value
```

**Workaround & Solution**
A short term workaround can be to avoid this feature. And for the parser the docker directives have to be on the first line of the file so it is valid for the whole file and can be found/detected quite good. So from what I see it might be one change to detect this directive and one to change the [regex](https://github.com/containerbuildsystem/dockerfile-parse/blob/master/dockerfile_parse/parser.py#L266) based on this detection.

schribl:
This change scans the first line if it is a comment for an escape directive as described in the docker [Documentation](https://docs.docker.com/engine/reference/builder/#escape). In case this is set the regex for the line continuation and the newly introduced state for the line continuation character is set. In addition the `_rstrip_backslash` was renamed to `_rstrip_eol` and takes one additional argument for the line continuation character.

Fixes #109 

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -239,11 +239,11 @@ class DockerfileParser(object):
              "value": "yum -y update && yum clean all"}
         ]
         """
-        def _rstrip_backslash(l):
-            l = l.rstrip()
-            if l.endswith('\\'):
-                return l[:-1]
-            return l
+        def _rstrip_eol(text, line_continuation_char='\\'):
+            text = text.rstrip()
+            if text.endswith(line_continuation_char):
+                return text[:-1]
+            return text
 
         def _create_instruction_dict(instruction=None, value=None):
             return {
@@ -262,15 +262,36 @@ class DockerfileParser(object):
 
         instructions = []
         lineno = -1
+        line_continuation_char = '\\'
         insnre = re.compile(r'^\s*(\S+)\s+(.*)$')  # matched group is insn
         contre = re.compile(r'^.*\\\s*$')          # line continues?
         commentre = re.compile(r'^\s*#')           # line is a comment?
+        directive_possible = True
+        # escape directive regex
+        escape_directive_re = re.compile(r'^\s*#\s*escape\s*=\s*(\\|`)\s*$', re.I)
+        # syntax directive regex
+        syntax_directive_re = re.compile(r'^\s*#\s*syntax\s*=\s*(.*)\s*$', re.I)
 
         in_continuation = False
         current_instruction = None
 
         for line in self.lines:
             lineno += 1
+
+            if directive_possible:
+                # once support for python versions before 3.8 is dropped use walrus operator
+                if escape_directive_re.match(line):
+                    # Do the matching twice if there is a directive to avoid doing the matching
+                    # for other lines
+                    match = escape_directive_re.match(line)
+                    line_continuation_char = match.group(1)
+                    contre = re.compile(r'^.*' + re.escape(match.group(1)) + r'\s*$')
+                elif syntax_directive_re.match(line):
+                    # Currently no information for the syntax directive is stored it is still
+                    # necessary to detect escape directives after a syntax directive
+                    pass
+                else:
+                    directive_possible = False
 
             # It is necessary to keep instructions and comment parsing separate,
             # as a multi-line instruction can be interjected with comments.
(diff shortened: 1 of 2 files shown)


MartinBasti:
Can you please update commit message with better description please? (That one you wrote to PR for example).

Otherwise looks great!

dockerfile_parse/parser.py
```diff
@@ -262,9 +262,11 @@ def _clean_comment_line(line):
 
         instructions = []
         lineno = -1
+        line_continuation_char = '\\'
         insnre = re.compile(r'^\s*(\S+)\s+(.*)$')  # matched group is insn
         contre = re.compile(r'^.*\\\s*$')          # line continues?
         commentre = re.compile(r'^\s*#')           # line is a comment?
+        directivere = re.compile(r'^\s*#\s*escape\s*=\s*(\\|`)\s*$', re.I)  # escape directive
```
MartinBasti, line 269: Dockerfile supports more (2 for now) directives, can we name it `escape_directive_re`?

dockerfile_parse/parser.py
```diff
@@ -275,6 +277,13 @@ def _clean_comment_line(line):
             # It is necessary to keep instructions and comment parsing separate,
             # as a multi-line instruction can be interjected with comments.
             if commentre.match(line):
+                # once support for python versions before 3.8 is dropped use walrus operator
+                if lineno == 0 and directivere.match(line):
```
MartinBasti, line 281: If I read docs correctly, `lineno == 0` is not always true. Directive can be on next lines if all previous lines are comments.

MartinBasti: Ahh sorry, I read it incorrectly

MartinBasti:
Uff there is an old pytest on RHEL7
```
============================= test session starts ==============================
platform linux2 -- Python 2.7.5 -- py-1.4.32 -- pytest-2.7.0 -- /usr/bin/python
rootdir: /builddir/build/BUILD/dockerfile-parse-1.0.0, inifile: 
collecting ... collected 0 items / 1 errors

==================================== ERRORS ====================================
____________________ ERROR collecting tests/test_parser.py _____________________
tests/test_parser.py:33: in <module>
    class TestDockerfileParser(object):
tests/test_parser.py:1356: in TestDockerfileParser
    pytest.param('\\', '`', marks=pytest.mark.xfail),
E   AttributeError: 'module' object has no attribute 'param'
=========================== 1 error in 0.22 seconds ============================
error: Bad exit status from /var/tmp/rpm-tmp.946Bu7 (%check)
```

dockerfile_parse/parser.py
schribl: I think you are right here, the current implementation will fail if there is first a `syntax` directive and in the next line a `escape` directive. But the docs support this case

schribl:
I pushed some changes:
* updated the logic to be more in line with the docker spec (syntax directive before escape directive)
* added a few more tests
* updated tests to work with pytest 2.7.x and latest pytest (they have incompatibles for `xfail` in `parameterize` so it now looks ugly)

MartinBasti (approved):
:rocket: Thank you!

ben-alkov (approved):
LGMT, thanks!

PR #119, opened 2021-06-03
Avoid adding repos to from scratch with alias

lkolacek:
Signed-off-by: Ladislav Kolacek <lkolacek@redhat.com>

CLOUDBLD-5945

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -767,7 +767,8 @@ class DockerfileParser(object):
         for stage in range(len(froms)-2, -1, -1):  # e.g. 0 for single or 2, 1, 0 for 3 stages
             start, finish = froms[stage], froms[stage+1]
             linenum = start['endline'] + 1 if at_start else finish['startline']
-            if skip_scratch and froms[stage]['value'] == 'scratch':
+            image, _ = image_from(froms[stage].get('value') or '')
+            if skip_scratch and image == 'scratch':
                 continue
             df_lines[linenum:linenum] = lines
 
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1176,6 +1176,8 @@ class TestDockerfileParser(object):
             ENV h i
             From scratch
             LABEL type=scratch2
+            From scratch as foo
+            LABEL type=scratch3
             """)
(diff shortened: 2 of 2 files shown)


dockerfile_parse/parser.py
```diff
@@ -767,7 +767,8 @@ def add_lines(self, *lines, **kwargs):
         for stage in range(len(froms)-2, -1, -1):  # e.g. 0 for single or 2, 1, 0 for 3 stages
             start, finish = froms[stage], froms[stage+1]
             linenum = start['endline'] + 1 if at_start else finish['startline']
-            if skip_scratch and froms[stage]['value'] == 'scratch':
+            if skip_scratch and (froms[stage]['value'] == 'scratch' or
+                                 froms[stage]['value'].lower().startswith('scratch as')):
```
MartinBasti, line 771: :( this makes me sad, so this parser doesn't recognize FROM value and stage name?

how about `scratch  as`, `scratch        as`, etc ...

MartinBasti: https://github.com/containerbuildsystem/dockerfile-parse/blob/master/dockerfile_parse/parser.py#L868 can we resue this to avoid duplicated logic?

lkolacek: I reused `image_from` PTAL.

dockerfile_parse/parser.py
```diff
@@ -767,7 +767,12 @@ def add_lines(self, *lines, **kwargs):
         for stage in range(len(froms)-2, -1, -1):  # e.g. 0 for single or 2, 1, 0 for 3 stages
             start, finish = froms[stage], froms[stage+1]
             linenum = start['endline'] + 1 if at_start else finish['startline']
-            if skip_scratch and froms[stage]['value'] == 'scratch':
+
+            from_line_value = froms[stage].get('value')
+            image = None
+            if from_line_value:
+                image, _ = image_from(from_line_value)
```
MartinBasti, line 774: Could be just
```python
image, _ = image_from(froms[stage].get('value') or '')
```
Is even possible that `value` is not defined?

lkolacek: yes we have that case in unit tests

MartinBasti (approved):

mkosiarc (approved):

rcerven (approved):

Issue #123, opened 2021-08-09
Option to only write data back to disk when explicitly requested


plannigan:
The parser currently will always write out the content to the file when a value is assigned to a property.  This is not immediately obvious, as it is only hinted at by the [docstring for fileobj](https://github.com/containerbuildsystem/dockerfile-parse/blob/9f22a64d8cd037cb191aca69dc36bba23a831aab/dockerfile_parse/parser.py#L100).

The [usage example](https://github.com/containerbuildsystem/dockerfile-parse#usage) doesn't mention anything about interacting with the file system, but will create or overwrite a file named `Dockerfile` file in the working directory.

If it would be to hard to adjust the library to work in a more explicit manor, it would be nice if the documentation were updated to be clear to someone finding the library for the first time about how it operates on files.

tim-vk:
While I was adding some other stuff to this repo I'd thought to pick up some other issues. This one is quite strange for me:

I agree the usage example is a bit minimal. But the python code itself is actually quite nicely documented:

```python
class DockerfileParser(object):
    def __init__(self, path=None,
                 cache_content=False,
                 env_replace=True,
                 parent_env=None,
                 fileobj=None,
                 build_args=None):
        """
        Initialize source of Dockerfile
        :param path: path to (directory with) Dockerfile; if not provided,
                     and fileobj is not provided, the current working
                     directory will be used
        :param cache_content: cache Dockerfile content inside DockerfileParser
        :param env_replace: return content with variables replaced
        :param parent_env: python dict of inherited env vars from parent image
        :param fileobj: seekable file-like object containing Dockerfile content
                        as bytes (will be truncated on write)
        :param build_args: python dict of build args used when building image
        """
```


The thing you are asking for (I think) is the `cache_content` input variable.
But there are two things that strike me as odd: 

1. why would you not want to edit the existing file? The only example I can think of is if you want to keep your original dockerfile and save an edited copy. But either way: you'd need two dockerfile parser instances (one to read, and one to edit/write). The `cache_content` variable seems useless to me.
2. As far as I can see: the `cache_content` variable doesn't work. It was added 7 years ago: My guess is something broke in the meantime. But honestly: I think it is clutter that should be removed.

But I am curious if there are any use cases that I'm missing since it was added with a reason.

Aisik00:
[...]

PR #126, opened 2021-12-15
Add flag autosave

even-wei:
A solution for #123 


even-wei:
Bad idea

tim-vk:
Why was this a bad idea? What would be a better solution?

PR #129, opened 2022-09-28
Adds tools for editing tags.

tim-vk:
Adds some functions for getting & setting image tags.
Adds a 'basetag' property for quickly adding/changing the tag of the baseimage.

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -401,7 +401,7 @@ class DockerfileParser(object):
             lines[instr['startline']:instr['endline']+1] = [instr['content']]
 
         self.lines = lines
-
+    
     @property
     def is_multistage(self):
         return len(self.parent_images) > 1
@@ -428,6 +428,21 @@ class DockerfileParser(object):
             raise RuntimeError('No stage defined to set base image on')
         images[-1] = new_image
         self.parent_images = images
+    
+    @property
+    def basetag(self):
+        """
+        :return: tag of base image, i.e. tag of base image
+        """
+        _, tag = tag_from(self.baseimage)
+        return tag
+    
+    @basetag.setter
+    def basetag(self, new_tag):
+        """
+        only change the tag of the final stage FROM instruction
+        """
+        self.baseimage = tag_to(self.baseimage, new_tag)
 
     @property
     def cmd(self):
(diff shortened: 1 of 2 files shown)


tim-vk:
@rcerven , @twaugh . Maybe a rude question but: is this project still alive? (Or am I just impatient ;) ).

rcerven:
> @rcerven , @twaugh . Maybe a rude question but: is this project still alive? (Or am I just impatient ;) ).

Hi, yes project is alive, we are just busy with osbs2

dockerfile_parse/parser.py
```diff
@@ -401,7 +401,7 @@ def parent_images(self, parents):
             lines[instr['startline']:instr['endline']+1] = [instr['content']]
 
         self.lines = lines
-
+
```
MartinBasti, line 404: This change shouldn't be here

MartinBasti (commented):
Hello, 

can you please describe describe the use case little bit more?

From my point of view, tag is tied with image, and not directly Dockerfile, for me tag is attribute of image and not direct dependency of dockerfile.

IMO better solution would be pass image string into a "ImageNameParser" and return tag form there, to keep responsibility separated.

dockerfile_parse/parser.py
MartinBasti: yes, dockerfile-parse works with imagename as a whole including digest or tag.

I'm not very eager to make it more granular, because we may also split imagename into: registry, namespace, image, tag/digest. There are multiple parts, many of them optional.

I'd like to keep dockerfile parse simple and rather use 3rd party library to manipulate with image string and just pass it to dockerfile parse

MartinBasti: We are using it like that, dockerfile just provides raw string, we manipulate it with separate code
https://github.com/containerbuildsystem/osbs-client/blob/master/osbs/utils/__init__.py#L372

MartinBasti: This is naming convention we use
```
    Naming Conventions
    ==================
    registry.somewhere/namespace/image_name:tag
    |-----------------|                          registry, reg_uri
                      |---------|                namespace
    |--------------------------------------|     repository
                      |--------------------|     image name
                                            |--| tag
                      |------------------------| image
    |------------------------------------------| image
```

tim-vk: Is there a specific 3rd party "image parser" that you know of that works with this convention?

I find it odd that osbs-client (your link) contains the logic to split up the image into registry, tag, name etc. which seems to be a bit counterintuitive with the "keep responsibility seperated" mentality.

The thing that I apparently am running against is that this repo doesn't seem to be a generic "dockerfile parser" but an "osbs dockerfile parser" which decreases the usability of it somewhat.

MartinBasti: I don't follow what do you mean.

What I want to avoid, is having separate method for everything: get_tag, set_tag, get_registry, set_registry, etc..

We are modifying registry part, we didn't put that code into docker_file parse, because it's better to have single class for manipulating with imagenames and then just write back string representation into dockerfile parse.

This also work for all stages in Dockerfile, not just final stage.

What could be good, is to modify dockerfile-parse to return imagename as object where you can use methods to modify image, or make this more object oriented, and return object per stage.

I'd like to keep here some OOP decoupling between image and dockerfile statements

tim-vk: I think I'm starting to get your intention. 

But if you'd truly want to decouple dockerfile statements I would expect something like `get_FROM_arguments()` Which would return all the possible arguments that are allowed for the `FROM` statement ([see dockerfile reference](https://docs.docker.com/engine/reference/builder/#from)). Or a class (`FromArguments`) where you can set the different arguments.

Regardless of the approach. What I don't really get is: Why isn't the `ImageName` class included in osbs-client? And not in dockerfile-parse? (or why isn't it a seperate repo?).

The thing thats bugging the most in the current set-up is that being able to change just the tag (or just the registry or other seperate component of "image") seems to be a fairly common & normal use case. And to achieve that either needs you to add your own functions (which I am sharing here) or install osbs-client (which doesn't seem to be available via pip?). Which raises my earlier question: Is `dockerfile-parse` meant as just a supplementary tool for osbs-client or as a tool for anyone who want to parse dockerfiles (and nothing else).

MartinBasti: > Why isn't the ImageName class included in osbs-client? And not in dockerfile-parse? (or why isn't it a seperate repo?).

Historical reasons, I don't know. Before it was part of atomic-reactor.


> Would it be more fitting for this project if the ImageName class ...
This would work for me, sounds like great idea. Let's wait what other maintainers things about it.

rcerven: moving ImageName here and returning base/parent images with it sounds good to me

tim-vk:
Since I'm using this specific branch for personal projects (and i don't want to update them all at once) I've created a new branch and PR: #139. 

Closing this PR.

PR #139, opened 2022-11-30
Move ImageName class from osbs client to utils of dockerfile parser.

tim-vk:
Summary of what is discussed in PR #129 :
- There are use cases only specific parts of an image (returned by baseimage) needs to be edited (for example: update all the tags).
- Currently dockerfile-parse does not provide any tools to perform this. A seperate "Image string parser" still needed to be added by any user of dockerfile-parse.
- [osbs client utils](https://github.com/containerbuildsystem/osbs-client/blob/master/osbs/utils/__init__.py#L372) already has a perfect solution: ImageName class with possibility to edit seperate components of the imagename string.
- ImageName() is a better fit for dockerfile-parse than for osbs-client.

This PR moves the ImageName class to utils of dockerfile-parse.

I did need to refactor the ImageName class slightly to get the following behavior in-line with dockerfile-parse behavior:
- Return None instead of empty strings.
- Do not automatically add "latest" if no tag is present.
- Improve __str__ behavior. (equality check)


tim-vk:
@rcerven : Is it the intention to keep this repo suitable for python 2? Just to be sure, before I'm going to put too much effort in trying to get that work. I'm having quite a bit of trouble wrapping my head around what's going wrong currently.

rcerven:
> @rcerven : Is it the intention to keep this repo suitable for python 2? Just to be sure, before I'm going to put too much effort in trying to get that work. I'm having quite a bit of trouble wrapping my head around what's going wrong currently.
> 
> It seems something with stringIO not accepting the result from the to_string() output. But not sure how i would be able to get it to work.

mmm for now afaik, @MartinBasti ?

MartinBasti:
> > @rcerven : Is it the intention to keep this repo suitable for python 2? Just to be sure, before I'm going to put too much effort in trying to get that work. I'm having quite a bit of trouble wrapping my head around what's going wrong currently.
> > It seems something with stringIO not accepting the result from the to_string() output. But not sure how i would be able to get it to work.
> 
> mmm for now afaik, @MartinBasti ?

We don't need it for py2 anymore, not sure about other community users :|

tim-vk:
@MartinBasti Whelp, i'm stumped: I've got everything up to snuff: Except for the python 2 bit... 
I have 0 clue how I would be able to fix these failing unit tests. Do you maybe have an idea?

Disabling the python2 tests is also a possibility. but that kind of feels like the cowards' way out...

MartinBasti:
> @MartinBasti Whelp, i'm stumped: I've got everything up to snuff: Except for the python 2 bit... I have 0 clue how I would be able to fix these failing unit tests. Do you maybe have an idea?
> 
> Disabling the python2 tests is also a possibility. but that kind of feels like the cowards' way out...

I'm really thinking about bumping major version and drop py2 support, py2 is dead and enterprise distros can use old versions of Dockerfile-parse, I'll check with other maintainers

MartinBasti:
> > @MartinBasti Whelp, i'm stumped: I've got everything up to snuff: Except for the python 2 bit... I have 0 clue how I would be able to fix these failing unit tests. Do you maybe have an idea?
> > Disabling the python2 tests is also a possibility. but that kind of feels like the cowards' way out...
> 
> I'm really thinking about bumping major version and drop py2 support, py2 is dead and enterprise distros can use old versions of Dockerfile-parse, I'll check with other maintainers

@tim-vk we agreed that we will do a major release of dockerfile-parse without py2 support. I'll drop py2 support in separate PR, so for now ignore py2 failures

dockerfile_parse/parser.py
```diff
@@ -880,7 +880,9 @@ def image_from(from_value):
         )?
         """)
     match = re.match(regex, from_value)
-    return match.group('image', 'name') if match else (None, None)
+    image = ImageName.parse(match.group('image')) if match else None
+    name = match.group('name') if match else None
+    return image, name
```
MartinBasti, line 885: we are breaking BW compatibility here, I'm not sure if we can do this

tim-vk: As far as i'm aware it shouldn't break backwards compatibility:

Old version: Image was a string
New version: Image is the new ImageName class which has a str representation (which is idential to the original string of the previous version).

This should be backwards compatible (the unit tests reflect this).

.packit.yaml
```diff
@@ -17,3 +17,7 @@ jobs:
     targets:
     - fedora-all
     - epel-8-x86_64
+
+srpm_build_deps:
```
tim-vk, line 21: This was apparently still missing in the master repo: I'm not 100% certain if this is correct. Please verify.

tests/test_parser.py
```diff
@@ -28,6 +28,116 @@
 instruction = instruction  # pylint: disable=self-assigning-variable
```
tim-vk, line 28: codeQL trips on this line. apparently you can ignore codeQL errors using an UI? (but i'm not sure where I would find this...).

tim-vk (commented):
move to 2 commits & finishing touches.

dockerfile_parse/parser.py
tim-vk: Addition:

Just realized what I typed above is not 100% true: See the explicit str(s) in util.py (line 51). 
impact should still be minimal.

Alternative solution would be keeping the old function.

MartinBasti: yes, explicit casting to str() is required, that's breaking change

MartinBasti: IMO would be safer to add new function and add deprecation warning to this one

tim-vk: Done!
I could not find any examples within the containerbuildsystem project with deprecations so i opted to use the `deprecation` module. Let me know if you want to use a different standard.

.packit.yaml
MartinBasti: I'm still not sure what this fix is fixing, we had not such issue previously

MartinBasti:
LGMT, very nice :)

rcerven:
can't you avoid making those breaking changes?  (osbs-client and atomic-reactor are using that) and I'd prefer to keep ImageName 100% backwards compatible

-    Return None instead of empty strings.
-    Do not automatically add "latest" if no tag is present.

tim-vk:
> can't you avoid making those breaking changes? (osbs-client and atomic-reactor are using that) and I'd prefer to keep ImageName 100% backwards compatible
> 
> * Return None instead of empty strings.
> * Do not automatically add "latest" if no tag is present.

While I recognize the wish for backwards compatibility, I do think that this is a great opportunity to genuinly think about "what do we want an imagename to be". And ignore the backwards compatibility bit (especially since the repo's you mention are not forced to use this version of the ImageName class. they can still use the old one. Though I hope someone will deprecate that one so everyone can switch to the new one with time).

So first: If there is nothing. What would you rather have: an empty string? or None. I would argue that None was a great choice to be returned. This makes checks a lot easier (`if(imagename.registry)`, is a beautiful way of checking if there is a registry).

Second: a dockerfile does not need to contain a `tag`. automatically filling it in with `latest` would be superflous (as per the [docker documentation](https://docs.docker.com/engine/reference/builder/#from) the builder does that for you).
additionally:  it would mean that if you would import a perfectly fine dockerfile & save it that the parser wil have changed it by adding 'latest'. I would find that undesired behavior.

Finally, If we do want to use the argument "backwards compatibility": The imageName class as implemented in this merge request is backwards compatible with existing code. e.g.: `image_from() == str(image_name_from())`.

So persionally I stand by these changes compared to the imageName class in the osbs-client. They seem to me to be the right way to go.

PR #144, opened 2023-09-28
Improve handling of file names and paths

enpaul:
Thank you so much for making this project! Each commit in this PR is self contained and could be reviewed separately, let me know if there's anything I can do to make the review process easier.

**Summary of changes:**

* Update `DockerfileParser` class to use `pathlib.Path` objects for file read/write operations
* Add `DockerfileParser.dockerfile` attribute to expose the path to the Dockerfile as a `pathlib.Path` object while maintaining backwards compatibility with the existing `DockerfileParser.dockerfile_path` property as a string
* Add `dockerfile_filename` keyword to `DockerfileParser.__init__()` to allow users to provide an alternative name for their dockerfile (ex: `Dockerfile.dev`, `Dockerfile.sample`, `Containerfile`, etc) while maintaining backwards compatibility with the existing default of `Dockerfile`
* Add tests for these new features

### DIFF
--- a/dockerfile_parse/parser.py
+++ b/dockerfile_parse/parser.py
@@ -9,8 +9,8 @@ of the BSD license. See the LICENSE file for details.
 
 import json
 import logging
-import os
 import re
+from pathlib import Path
 from contextlib import contextmanager
 from shlex import quote
 
@@ -100,6 +100,7 @@ class DockerfileParser(object):
         """
 
         self.fileobj = fileobj
+        self.dockerfile = None
 
         if self.fileobj is not None:
             if path is not None:
@@ -107,11 +108,11 @@ class DockerfileParser(object):
             else:
                 self.fileobj.seek(0)
         else:
-            path = path or '.'
-            if path.endswith(DOCKERFILE_FILENAME):
-                self.dockerfile_path = path
+            path = Path(path) or Path.cwd()
+            if path.name.endswith(DOCKERFILE_FILENAME):
+                self.dockerfile = path
             else:
-                self.dockerfile_path = os.path.join(path, DOCKERFILE_FILENAME)
+                self.dockerfile = path / DOCKERFILE_FILENAME
 
         self.cache_content = cache_content
         self.cached_content = ''  # unicode string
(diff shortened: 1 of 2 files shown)


tests/test_parser.py
```diff
@@ -1509,3 +1509,11 @@ def test_escape_directive_ignore_after_comment(self, dfparser):
...
             }
         ]
+
+    def test_alt_dockerfile_names(self, tmpdir):
+        content = "FROM fedora:38"
+        out = DockerfileParser(path=tmpdir, dockerfile_filename="Containerfile")
+        out.content = content
+
+        validate = DockerfileParser(path=tmpdir, dockerfile_filename="Containerfile")
+        assert validate.baseimage == out.baseimage
```
chmeliik: Could this also test the existence of `tmpdir / "Containerfile"`? Technically, this test would pass even if dockerfile_filename didn't do anything

tests/test_parser.py
```diff
@@ -1517,3 +1517,13 @@ def test_alt_dockerfile_names(self, tmpdir):
 
         validate = DockerfileParser(path=tmpdir, dockerfile_filename="Containerfile")
         assert validate.baseimage == out.baseimage
+
+    def test_dockerfile_path_compatibility(self, tmpdir):
+        parser = DockerfileParser(path=tmpdir)
+        assert str(parser.dockerfile) == parser.dockerfile_path
+        assert parser.dockerfile == tmpdir / "Dockerfile"
```
chmeliik, line 1530: nitpick: `tmpdir` is not a pathlib.Path, it is a py.path.local https://docs.pytest.org/en/7.4.x/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures

It looks like Paths and py.paths are considered equal though, so the test still works

dockerfile_parse/parser.py
```diff
@@ -149,9 +151,16 @@ def _open_dockerfile(self, mode):
...
-            with open(self.dockerfile_path, mode) as dockerfile:
+            with self.dockerfile.open(mode) as dockerfile:
                 yield dockerfile
 
+    @property
+    def dockerfile_path(self):
+        """
+        :return: Path to the Dockerfile as a string, or None if using fileobj
+        """
+        return str(self.dockerfile) if self.dockerfile else None
```
chmeliik, line 164: I think it was previously possible to set `dockerfile_path` after the object was already created. It wasn't necessarily an _intended_ use case, but to avoid breaking changes, could `dockerfile_path` also have a setter?

chmeliik (commented):
Thanks for the contribution!

Looks good to me, just one backwards compatibility consideration and some unittest nitpicks

dockerfile_parse/parser.py
```diff
@@ -149,9 +153,20 @@ def _open_dockerfile(self, mode):
...
+
```
chmeliik: flake8 is saying there's extra whitespace on this line (165)

chmeliik (approved):
LGTM except 1 linter error

the pylint and markdownlint errors are unrelated

dockerfile_parse/parser.py
enpaul: Ah sorry, missed that somehow.

